PR #13, opened 2025-05-29
Refactor miner

n13:
## Refactor into miner and wallet 

Made 3 packages out of the wallet

- quantus_sdk for any app to use our backend services
- mobile_app for the existing mobile app - now using quantus_sdk 
- app for the miner app (TBD)


illuzen:
LGTM

PR #31, opened 2025-07-08
Add reversible transaction events

n13:
- Refactor main screen with only reversible events that are live
- Added tx history screen
- Fix layouts
- Design for the all transactions screen

Still missing:
- Cancel transactions by hitting the cancel button
- All transactions screen filter

Main Screen

Simulator Screenshot - iPhone 16 Pro - 2025-07-08 at 18 04 10

All Transactions

Simulator Screenshot - iPhone 16 Pro - 2025-07-08 at 20 07 47


dewabisma (approved):
Insanely huge PR! 

Looking great overall. I just wondering if the private function in transaction list item like `_formatAddress` and the other only used there?

PR #34, opened 2025-07-10
Formatting cleanup

n13:
Formatting main number with dividers like 1,000 (instead of 1000)

Updated tests

Clean up address formatting, added address formatting service


dewabisma (approved):
LGTM!

PR #48, opened 2025-07-14
Wallet state manager 2

### LINKED ISSUE #33: Transaction State Machine
Problem:

Right now we wait for transactions to be included in a block - this takes a while

We want all UX operations to take at most 0.4 seconds. 

Solution: Sending transactions needs to be asynchronous. 

Problem: Things like balance and subsquid take a while to update - on sending, balance on chain is only updated once a block has been processed, ie 10+ seconds. Subsquid will take a little longer to catch up. 

Solution: State machine queues transaction types such as transfer, reversibleTransfer

We can use the same types we already have but add local state to them

State can be: 
Created
Sent (to chain)
Included in block
Executed

There may be more of these. Substrate and polkadart use a ws connection so we can track the status of the extrinsic, and we get different notifications. 

The solution will be to run the extrinsic in a background thread and update the state machine as needed. 

There could be multiple simultaneous transactions "in flight" and when we update the balance from chain, we need to figure out which of our transactions influence the balance (lets say the user hits refresh on the main screen). 

Transaction state machine will provide a list of transactions that have been created but not yet executed. 

They can be in either of these states, but they can also fail at each state too

Example - can't submit to node, submitted but not executed, etc. 

### Send reversible with 10 minutes 
1 - Hit send button -> Sending
2 - Sent to chain complete -> Pending on chain
3 - Balance changes on chain -> Added to scheduler on chain
4 - Event appears in subsquid (history list, small delay)
... 10 minutes later... 
5 - Timer hits 0
6 - Tx is now pending on chain (meaning any time now)
7 - Tx executed (balances change, frozen amount is unfrozen, subtracted from the free balance, and added to the target balance)

### Send Transfer
1 - Hit send button
2 - Sending... (takes some seconds)
3 - Pending on chain (takes block time seconds, 10 or 20, but can vary)
4 - Executed on chain

### Receive

Receive looks at the same thing from the other side, however, balances don't change for receive until receive is complete, so that's ok, we don't need to keep those in our list. 

You

n13:
- Ephemeral balances and transactions
- Handling state changes of tx in background
- Polling history and balance once tx is in block
- Cancel stream subscription after inBlock
- Handle reversible transfers

Cleanup tasks added as issues since this is already so big:
#49 
#50 
Load more transfers on scroll on tx screen - this could be a different PR -> use different PR for this. See #51 

Fix #33

### DIFF
--- a/miner-app/lib/features/miner/miner_dashboard_screen.dart
+++ b/miner-app/lib/features/miner/miner_dashboard_screen.dart
@@ -71,7 +71,6 @@ class _MinerDashboardScreenState extends State<MinerDashboardScreen> {
           _walletBalance = 'Address not set';
           _walletAddress = null;
         });
-        // TODO: Implement navigation to rewards address setup screen
         print('Rewards address mnemonic not found. Redirecting to setup...');
         // Example Navigation (requires go_router setup)
         // context.go('/rewards_address_setup');
@@ -81,7 +80,6 @@ class _MinerDashboardScreenState extends State<MinerDashboardScreen> {
         _walletBalance = 'Error fetching balance';
         _walletAddress = address;
       });
-      // TODO: Show a more user-friendly error message (e.g., Snackbar)
       print('Error fetching wallet balance: $e');
     }
   }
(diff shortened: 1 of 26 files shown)


dewabisma (approved):
Looking great! I don't see any obvious flaw yet logically. I wanted to check more actually for readability but I can't seem get clear look on how readable from this PR. So, I will just do that when I read the codebase to see if there are some confusing parts that need refactor or more comments.

Issue #54, opened 2025-07-16
Send overlay design tweaks


n13:
No changes to function but it looks different in the new design. 

I don't think the last screen has the right text in the design, and also cannot have a revert button under it, since we can't even revert it at the time. 

Revert needs to be done from the main UX. 

So no revert button. We can always change the title wording no big deal. 

<img width="575" height="432" alt="Image" src="https://github.com/user-attachments/assets/678d2335-f7c9-4366-a7a0-a61bafa3a0ca" />

PR #58, opened 2025-07-17
Update send overlay and update receive overlay backdrop

dewabisma:
Issue related #54 

I have updated most of it, the only missing implementation is getting hash so the url will be correct. I don't understand how we can get the hash when the state of the sending is complete. From the code, it seems the status will be pending before we can get the hash. 

Below are the screenshots of the updated UI

1. Sending confirm
image

2. Sending loading
image

3. Sending complete
image

4. Opening explorer
image

I also update the receiving screen to have purple tint color
image


n13 (commented):
You have line length on 80, I have 120

but I like 80 more now, so will make another PR to enforce 80

n13:
I don't think we can jump to explorer since the tx won't be in the explorer for a while

Also should say "Sending" at the last screen since it also works for instant transactions ... it should also not show the remaining time when reversible time is set to zero. Needs to work for both cases.

n13:
Ok lets hit it I will change text later.

n13 (approved):

PR #59, opened 2025-07-18
set line width to 80

n13:
Change line width to 80, add linter enforcement.

No code logic changes, just formatting


dewabisma (approved):
Approved!

PR #60, opened 2025-07-18
remove special treatment of index 0 to be compatible with other wallets

n13:
note this small change will change all account keys, so all keys will be different from now on also fixed some margins


dewabisma (approved):
LGTM!

Issue #43, opened 2025-07-10
Detailed Info Action Pane for all transactions


n13:
When a user taps a transaction (other than currently scheduled reversibles), there's a detail screen with block explorer link that pops up

See figma for more 

<img width="209" height="800" alt="Image" src="https://github.com/user-attachments/assets/ae8ba23c-884e-4dfc-9aa3-b7dd07250e56" />

PR #68, opened 2025-07-18
Update non scheduled transaction detailed info action sheet

dewabisma:
Related issue #43

I have implemented for all except scheduled tx. The only missing piece is the retry button logic on failed tx.
I'm not sure how to handle that yet.

image


n13 (approved):
Would advise to not use external packages unless really needed

Otherwise we collect packages, which are hard to maintain. 

For dotted border we can just copy the code.

PR #69, opened 2025-07-19
Lint fixes and details

n13:
- removed dotted border import
- fixed all warnings (mostly line length)
- changed some copy actions back to account ID

The benefit of only copying ID is someone else can just take it and paste it and it works.


dewabisma (approved):
Great, so it will check and fix formatting or analyzer error?

PR #70, opened 2025-07-20
Local auth

n13:


n13:
> * Why do we ask for biometric auth on enabling the local auth while when disabling we don't need to? In my opinion this is weird, so anyone can disable the local auth then.

That's my bad - that was in my instructions to the AI, I thought we don't need to do that. 

But, generally, disabling a security feature requires authentication so you're right. 

I changed the code to make it so.

dewabisma (approved):
All right, looking great!

Issue #71, opened 2025-07-21
Sending amount value formatting/validation


dewabisma:
### Formatting
When defining amount to send in send screen. We can have 0 followed by other number instead of enforcing comma after 0. 

Expected amount format,
0.5 QUAN, 0.1 QUAN, 1 QUAN

What still possible and shouldn't be allowed to be inputted,
05 QUAN, 01 QUAN, 001 QUAN, 0000000 QUAN

### Large number handling
If we enter a really big sum of QUAN, it will not be readable anymore since it will overflow the input constraint.

### Decimal internationalization
The input has regex that only accept dot(.), while some country use comma(,) to define decimal separator. 

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/90953c64-d6bf-420b-b1e7-7b38f83f0906" />

PR #74, opened 2025-07-21
Refactor reusable widget + input decimal formatting 

dewabisma:
Related issue #71 

Other than fixing the formatting issue in send screen. I also start doing refactor little by little. Please give me your thought on this, if I should do it for all of our widgets gradually.


mobile-app/lib/shared/utilities/decimal_formatter.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:flutter/services.dart';
```
n13, line 1: Normally I put extensions in /extensions... might be better than utilities

quantus_sdk/lib/src/services/number_formatting_service.dart
```diff
@@ -67,12 +72,16 @@ class NumberFormattingService {
...
-      final decimalAmount = Decimal.parse(formattedAmount);
+      final sanitizedText = formattedAmount.replaceAll(',', '.');
+
+      final decimalAmount = Decimal.parse(sanitizedText);
       // Check if input precision exceeds chain precision
       if (decimalAmount.scale > decimals) {
         // Option 1: Truncate (like toBigInt does)
         // Option 2: Throw an error - let's stick with truncation for now
-        debugPrint('Warning: Input amount $formattedAmount exceeds $decimals decimals, will be truncated.');
+        debugPrint(
```
n13, line 82: we should now allow any entries that are an illegal value. So we only allow up to 12 decimals.

mobile-app/lib/shared/utilities/decimal_formatter.dart
dewabisma: Okay! Just me doing it like in web thingy 🤣

thought it’s not really extending, do we put in extrnsions? I thought extension is when we add more func to built in class

dewabisma:
> I've intentionally gone very light on trying to put screens in sub folders.
> 
> For send it's OK though since it's a few pieces.
> 
> The extension should be in extensions not utilities (extensions folder basically being the same as utilities but for extensions. Reading this code also I think it makes sense because when you don't know the codebase it's helpful when extensions are separated from normal classes.

I think colocating help us easily navigate it. Especially for someone new, and separating class also in case it’s not tightly related. I wanted to clean more each widget, like making reusable component out of common pattern.

mobile-app/lib/shared/utilities/decimal_formatter.dart
n13: ...

class DecimalInputFormatter extends TextInputFormatter

seems to be extending ...

n13:
image
there's already extensions... you could move it to quantus sdk... we will probably need it for the miner

PR #79, opened 2025-07-22
Fix auth wording

n13:
minor issue it would ask the user for enabling even if they were disabling - you'd only see this text on the pin lock screen in case biometrics fail.


dewabisma (approved):
Okay!

PR #84, opened 2025-07-23
Cleaning up polkadart

n13:
We need to delete old code before generating - generating code doesn't delete existing items so it just keeps adding and never deleted outdated or no longer existing metadata classes! Fixing this by adding a delete to the scripts. Also deleted all "polkadot" chain code, we don't need this.


dewabisma (approved):
Nice, save a huge amount of space!

Issue #86, opened 2025-07-23
Received / Detail popup is wrong for reversible transaction


n13:
Steps 

1. Tap on a reversible transaction to see details

Expected result:
Detail overview indicates the transaction can be reversed - showing timer, date of fulfillment etc. 
Title: Receiving (not "received")
Receiving in ....<time>... (hours:minutes:seconds)

Actual:
Detail overview looks like a normal already completed transfer

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/2bf59f80-8c7c-4ab9-a589-a0eba68c6174" />

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/6266f864-6b3e-4e11-846d-b4dcb2b4b713" />

PR #92, opened 2025-07-24
Reversible transaction details action sheet

dewabisma:
Related issue #86 

Fixed the reversible sheet, also fix styling on tx listing item, some minor UI fixes.

- Homepage listing item
image

- Reversible sending tx details
image

- Reversible receiving tx details
image

### DIFF
--- a/quantus_sdk/lib/src/services/datetime_formatting_service.dart
+++ b/quantus_sdk/lib/src/services/datetime_formatting_service.dart
@@ -42,8 +42,8 @@ class DatetimeFormattingService {
     return '${_formatPast(difference)} $timeString';
   }
 
-  static FormattedDuration formatDuration(Duration? duration) {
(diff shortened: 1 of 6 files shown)


quantus_sdk/lib/src/services/datetime_formatting_service.dart
```diff
@@ -23,6 +42,35 @@ class DatetimeFormattingService {
     return '${_formatPast(difference)} $timeString';
   }
 
+  static FormattedDuration formatDuration(Duration? duration) {
+    if (duration == null || duration.isNegative) {
```
n13, line 46: Are we ever expecting to pass in a null duration here? 

Seems we should crash when this happens, rather than show 00:00:00

n13 (changes requested):
Requesting one change

Everything else looks great.

n13 (approved):

PR #101, opened 2025-07-24
Retry transactions a few times - fixes #96

### LINKED ISSUE #96: Send fails too fast.
Make sure it retries 3 times before showing a user facing error

This used to work, that's a regression. 

I got a send failure much too fast, retried, and it worked. 

So I assume it no longer retries.

n13:
Fix for #96 

Human Error: When implementing async wallet, I accidentally broke the retry logic, as we return a stream future now. 

Fix:Handle errors where they occur


dewabisma (approved):
Seems good!👍🏽 Just need to fix the analyzer error

PR #102, opened 2025-07-24
remove failed ephemeral transactions

### LINKED ISSUE #98: Remove failed transactions from history
There's currently too many ways they can fail and it's no point to show them either for sender or receiver. 

Image

n13:
otherwise they just pile up there for no reason. Fixes #98


dewabisma (approved):
Great!

PR #103, opened 2025-07-24
Wallet state manager update

### LINKED ISSUE #94: Tx don't always update from inBlock to countdown timer
Chris reported transition from inBlock to timer required pull to refresh. 

Maybe there was a network error along the way

Seems to happen when sending multiple. 

Regression:
I think because they don't have a unique ID yet ... the widget keys are then not unique. Need to debug this.

n13:
This might fix the stuck pending updates. 

- Quiet background updates don't cause error overlay to appear
- Only 1 history poll timer running at one time. 

The timer issue wasn't really a bug but if you had say 5 items pending it would end up creating 5 timers, increasing the chance for failure to load and increasing server load. 

We only need one polling timer run at any given time. 

Fixes #94

### DIFF
--- a/mobile-app/lib/models/wallet_state_manager.dart
+++ b/mobile-app/lib/models/wallet_state_manager.dart
@@ -70,6 +70,9 @@ class WalletStateManager with ChangeNotifier {
   @override
   void dispose() {
     _pollingTimer?.cancel();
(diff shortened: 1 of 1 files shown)


n13:
Merging since it fixes some serious bugs.

PR #107, opened 2025-07-25
countdown timer is less dumb

n13:
it may come back from background!


n13:
I unified duplicate code everywhere.

Issue #113, opened 2025-07-25
Wallet list should not push create new button


dewabisma:
<img width="1179" height="2556" alt="Image" src="https://github.com/user-attachments/assets/5de3e7b4-2b46-411b-8f65-c3749fcc3478" />

PR #122, opened 2025-07-25
Fix create account btn pushed

dewabisma:
Related issue #113 

Now it's not pushing down the button.

image


n13 (approved):
if it works, it works

PR #123, opened 2025-07-25
Fix navbar item hit box

dewabisma:
Now it should be more easier to tap the nav item


dewabisma:
Yup, it looks the same, just wider hit box

PR #124, opened 2025-07-25
Clean up network fee handling

n13:
Minor change but prettytrsubtle behavior

- Send is disabled until we have the network fee
- We assume any amount has the same network fee, so we only load it once
- We don't set to zero on validation errors - user is still making a send transaction.


dewabisma (approved):
LGTM!

PR #131, opened 2025-07-28
remove duplicate and unused code, improve efficiency

n13:
we no longer pull the entire block on get fee and send, only the header - all we need is the block number also pulling all values in a single future.


dewabisma (approved):
LGTM!

PR #132, opened 2025-07-29
Add account discovery

n13:
## Overview

This PR solves the issue that miner accounts can't see what's in their wallet

## Solution
Added a feature we need anyway, account discovery. When a user imports a wallet, we check if there's any existing wallets with balances based on HD key derivation. 

## Limitations

Currently only checks the first 20, we can easily add more later, or have this in settings to search for unlimited accounts. 

Or else users who just keep creating new accounts using the create button, will also find their own accounts back one by one. 

## Code changes

- Account UI position prep for allowing users to move accounts around. Right now accounts are sorted by index, but index is the HD generation key so it's an internal value that can't be changed. 
- Fix up unit tests to align with CLI generated old accounts - make sure we generate the same account addresses as CLI from mnemonic. 
- Add old account method to account discovery

### DIFF
--- a/miner-app/lib/features/miner/miner_dashboard_screen.dart
+++ b/miner-app/lib/features/miner/miner_dashboard_screen.dart
@@ -53,7 +53,7 @@ class _MinerDashboardScreenState extends State<MinerDashboardScreen> {
       if (mnemonic != null) {
         // Derive keypair from mnemonic using SubstrateService (exported by quantus_sdk)
         // ignore: deprecated_member_use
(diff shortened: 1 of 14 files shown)


dewabisma (approved):
Cool! Looking good, just I don't understand what uiPosition is for in Account model.

n13:
> Cool! Looking good, just I don't understand what uiPosition is for in Account model.

The position in the list. 

We will need it when we allow people to re-arrange the list ;) 

I guess though maybe this is only one way to do it; we might just as well store a separate UI list - which is probably better. 

Let me remove this again ;)

Issue #119, opened 2025-07-25
Account setting minor style fix


dewabisma:

dewabisma:
Need to work more on small device

PR #134, opened 2025-07-30
UI fix for iPad + few bug fixes

dewabisma:
Related issue #119 #130 #128 

I have finished updating the app to be compatible with iPad. Haven't tested on android tablet but if they have similar screen size then should be fine. I also did some refactor by extracting repeated pattern into reusable function or method or extensions.

I have also use the `WalletAppBar` component for all screen that need it so we have consistent app bar. Changed `Checkbox` to `CheckboxListTile` for improvement in accessibility.

Below are some screenshots of the iPad view:

image

image

image

image

image


mobile-app/lib/features/components/recent_address_list_item.dart
```diff
@@ -76,7 +79,7 @@ class _RecentAddressListItemState extends State<RecentAddressListItem> {
             widget.address,
             style: TextStyle(
               color: Colors.white.useOpacity(0.60),
-              fontSize: 11,
+              fontSize: isTablet ? 16 : 11,
```
n13, line 82: this would be a lot better with a global call - get me the font size 11 for example, and the global function would check isTablet and return 16 or 11 depending. Something like that. 

It's not good to pass this around all over the place. 

Could also check what the common ways are that this is solved. 

We should not have if statements all over the codebase. If it's in one or two locations, that's ok, else it should be in dedicated classes that abstract these things out.

n13:
CI is failing and also conflicts.

Issue #135, opened 2025-07-31
Transfer max not working


n13:
Steps 
- send
- hit max - send entire balance 
- confirm send 

Expected: entire balance is sent 
Actual: error 

This is reported by a user has to do with fee and maybe also keep alive vs not keep alive / existential deposit

dewabisma:
Solved #144

PR #137, opened 2025-07-31
Fix transfer max amount

dewabisma:
Related issue #135 

So the reason it's not working is because the network fee is 0 at the moment it trying to calculate max amount, so I made the return the network fee that is fetched.

### DIFF
--- a/mobile-app/lib/features/main/screens/send/send_screen.dart
+++ b/mobile-app/lib/features/main/screens/send/send_screen.dart
@@ -25,7 +25,7 @@ class SendScreenState extends State<SendScreen> {
   final NumberFormattingService _formattingService = NumberFormattingService();
   final SettingsService _settingsService = SettingsService();
   BigInt _maxBalance = BigInt.zero;
-  BigInt _networkFee = BigInt.zero; // Actual network fee fetched from chain
+  BigInt? _networkFee; // Actual network fee fetched from chain
   bool _isFetchingFee = false;
   BigInt _amount = BigInt.zero;
   bool _hasAddressError = false;
@@ -38,14 +38,14 @@ class SendScreenState extends State<SendScreen> {
 
   late Future<BigInt> _balanceFuture;
 
+  bool get _hasBalance => _maxBalance <= BigInt.zero;
+  bool get _haveNetworkFee => _networkFee != null;
+
   @override
   void initState() {
     super.initState();
     _balanceFuture = _loadBalance();
     _loadReversibleTimeSetting();
-    // Listen for changes in recipient and amount to update fee
-    _recipientController.addListener(_debounceFetchFee);
-    _amountController.addListener(_debounceFetchFee);
   }
 
   @override
@@ -94,10 +94,8 @@ class SendScreenState extends State<SendScreen> {
   }
 
   Future<void> _lookupIdentity() async {
-    if (!mounted) return; // Add early return if not mounted
     final recipient = _recipientController.text.trim();
     if (recipient.isEmpty) {
-      if (!mounted) return; // Check mounted before setState
       setState(() {
         _savedAddressesLabel = '';
         _hasAddressError = false;
@@ -107,7 +105,6 @@ class SendScreenState extends State<SendScreen> {
 
     try {
       final isValid = _isValidSS58Address(recipient);
-      if (!mounted) return; // Check mounted before setState
       setState(() {
         _hasAddressError = !isValid;
       });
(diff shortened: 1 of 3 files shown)


n13:
We already disable send until we have the fee - why not just disable "max" until we have the fee. 

Also remove the code that sets fee to 0 when it doesn't exist - it should just crash when there's no fee. Crash early would have saved us here ;) sloppy AI code (which I added BTW).

dewabisma:
> We already disable send until we have the fee - why not just disable "max" until we have the fee.
> 

Well, they are working differently. Because max can be pressed in the beginning on when after user type any amount. And, fee is calculated after user type not when user press max. So gas fee will be 0 at that moment when we wanted to put the amount. The other solution will be to let them click max and waiting for gas fee fetched then we can update the amount with the correct subtracted gas fee.

dewabisma:
> Also remove the code that sets fee to 0 when it doesn't exist - it should just crash when there's no fee. Crash early would have saved us here ;) sloppy AI code (which I added BTW).


Which one do you refer it to? if you are referring this snippet,
```dart
Future<BigInt> _fetchNetworkFee({bool isMaxSend = false}) async {
  final recipient = _recipientController.text.trim();
    final isInvalidMaxSend = isMaxSend && _maxBalance <= BigInt.zero;
    final isInvalidAmount = _amount <= BigInt.zero && !isMaxSend;
    final isInvalidAddress = !_isValidSS58Address(recipient);

    if (isInvalidAddress ||
        isInvalidAmount ||
        isInvalidMaxSend ||
        _haveNetworkFee) {
      setState(() {
        _networkFee = _networkFee;
        _isFetchingFee = false;
        _hasAmountError =
            _amount > BigInt.zero && (_amount + _networkFee) > _maxBalance;
      });
      return _networkFee;
    }
    setState(() {
      _isFetchingFee = true;
    });
    ...
```

Then I don't believe we should crash here, why? We should just say error to user for invalid address or when the balance is insufficient. We should set the _networkFee to null though since we don't get the network fee because of invalid input.

`isInvalidAmount` can happen when user type 0 or bigger number than he/she could type. We might want to automatically cap user input to their max balance - fee. And that `isInvalidMaxBalance` yeah, actually I think we can simplify it by making sure user has any balance, otherwise we disabled the max button.

n13:
Ok since we got new information, the fee is not actually fixed... 

We need to do the fees fetching again after amount is entered, and max value button needs to do inline fee calculation in order to find the correct value that works. 

We need to also change our substrate service fee calculator since we can't use a dummy value, we need to use the real value. 

In a follow up PR I will fix the substrate service so that fee calculation also work for reversible transactions - pretty sure that won't work now because it uses the scheduler which will add feees.

mobile-app/lib/features/main/screens/send/send_screen.dart
```diff
@@ -38,14 +38,14 @@ class SendScreenState extends State<SendScreen> {
...
+
   @override
   void initState() {
     super.initState();
     _balanceFuture = _loadBalance();
     _loadReversibleTimeSetting();
-    // Listen for changes in recipient and amount to update fee
-    _recipientController.addListener(_debounceFetchFee);
-    _amountController.addListener(_debounceFetchFee);
   }
```
n13, line 49: why were these removed?

mobile-app/lib/features/main/screens/send/send_screen.dart
```diff
@@ -219,14 +217,17 @@ class SendScreenState extends State<SendScreen> {
...
       });
+
+      return estimatedFee;
     } catch (e) {
       print('Error fetching network fee: $e');
       setState(() {
         _isFetchingFee = false;
-        _hasAmountError = _amount <= BigInt.zero || _amount > _maxBalance;
+        _hasAmountError = _amount > _maxBalance;
       });
```
n13, line 229: Why was this changed?

mobile-app/lib/features/main/screens/send/send_screen.dart
```diff
@@ -219,14 +217,17 @@ class SendScreenState extends State<SendScreen> {
...
         _isFetchingFee = false;
-        _hasAmountError = (_amount + _networkFee) > _maxBalance;
+        _hasAmountError = (_amount + estimatedFee) > _maxBalance;
       });
+
+      return estimatedFee;
     } catch (e) {
       print('Error fetching network fee: $e');
       setState(() {
         _isFetchingFee = false;
```
n13, line 227: Apart from it being changed, kind of looks wrong anyway - we just ignore network fee here... hmm.

n13 (changes requested):
See comments

mobile-app/lib/features/main/screens/send/send_screen.dart
dewabisma: Because we can’t send it on empty amount anyway so no need to check amount <= zero

quantus_sdk/lib/src/services/number_formatting_service.dart
```diff
@@ -67,14 +69,15 @@ class NumberFormattingService {
   ///
   /// Returns null if the input string is invalid.
   BigInt? parseAmount(String formattedAmount) {
-    if (formattedAmount.isEmpty) {
-      return BigInt.zero;
-    }
+    if (formattedAmount.isEmpty) return null;
 
     try {
-      final sanitizedText = formattedAmount.replaceAll(',', '.');
```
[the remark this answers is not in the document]
dewabisma: There is bug about parsing because of locale difference. It fix that bug. More solid now using the intl library.

Okay I will add test case later

PR #140, opened 2025-08-01
Core Architecture Refactor

n13:
Using riverpod to manage transactions state

Sane way of managing 
- Pending events
- History events
- Reversible events
- The state transitions between these
- Slow history polling
- Balance updates
- Effect of pending transactions on balance. 
- Invalid transaction retry

This is already massive but there's some more things to do to go in new PRs: 

- Test invalid transaciton retry by say providing an invalid nonce - set up test cases to make the 'invalid' transactions happen. 
- Change reversible polling service to set timers that expire exactly when the transactions expire, and start poling then, instead of trying to closely monitor the timer for 30 seconds each time
- Combine reversible pollers into one timer rather than having one timer per reversible - in case they overlap although if we change it so we only start polling at zero that won't be needed, would be over-optimized. 
- Test account updates - create new account and so on
- Test incoming transfers thoroughly 
- Implement all transactions screen from the PR with these new providers.


dewabisma (approved):
Wonderful! The state management is much more granular now, really love it, 

I found two things that I wanted to comment.

1. Do we want to keep tx tracking and reversible polling in background even if app not focused?
My point is that, we can do a fresh fetch when app come to foreground(focused/resumed), so is there any need to keep the polling active in background? It's will make the app consume much resource (battery, bandwidth).

2. Pagination has more seems not reliable
We are still using `items.length == _limit` which in my opinion is not bullet proof for checking whether we can fetch more. Ideally, we should have a total available count. I think GraphQL has it. I use it in website to do pagination.

PR #144, opened 2025-08-01
Fix max button

n13:
- Added get fee to max button function, like #137 
- No thousands separators in our send field
- Fetch network fee when switching from / to reversible / non-reversible - since they have different fees
- Show error when no valid recipient
- Show error when fee fetch fails
- Factor out getNetworkFeeForAmount
- Send max sets the amount directly - factored out _setSendAmount for this purpose


dewabisma (approved):
Looking good to me, just that like I mentioned in my old pr. The addListener for controller in override init is redundant, because we already call it on change of each input.

PR #146, opened 2025-08-04
protect routes with auth wrapper

n13:
- need to test this..


n13:
I have a much simpler fix in another branch, closing

Issue #158, opened 2025-08-05
Transfer max for reversible transfer stuck at inBlock (failling)


dewabisma:
## The problem
When I try to transfer max to another account using reversible transfer it stuck at inBlock pending status.

## Expected
It should proceed to scheduled and has timer.

## How to reproduce
1. Go to send screen
2. Enter address
3. Tap on max transfer button
4. Set reversible time 
5. Tap send
6. Confirm transfer

## Screenshot

<img width="1668" height="2420" alt="Image" src="https://github.com/user-attachments/assets/bd260307-7914-49e0-b7a3-0d366533c4a6" />

n13:
Fixed #160

PR #160, opened 2025-08-05
Fix send max reversible

n13:
Fixed 2 bugs:
- Failed transactions would not be removed correctly from pending
- Reversible max was not working #158 

First fix: Now polling block hash for pending transactions - this way we know for sure whether or not the transaction was successful. If we find the block, we check if it's in there, if yes, it worked, if no, it failed.

Second fix: Reversible needs to subtract existential deposit from max send. Else it won't work. 

User feedback could be better since it's not actually a fee for existential deposit. We should pop something saying you can't send the entire balance reversible, only balance - existential fee. 

Also there is then 0.001 left in the account, not 0. 

But unless we change the runtime, that's how it works.

Could be that we can't freeze more assets also. Not sure where the existential deposit check happens.

### DIFF
--- a/mobile-app/lib/services/reversible_transfer_monitoring_service.dart
+++ b/mobile-app/lib/services/reversible_transfer_monitoring_service.dart
@@ -1,7 +1,9 @@
 import 'dart:async';
 
+import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:quantus_sdk/quantus_sdk.dart';
+import 'package:resonance_network_wallet/app_lifecycle_manager.dart';
 import 'package:resonance_network_wallet/providers/all_transactions_provider.dart';
 import 'package:resonance_network_wallet/providers/wallet_providers.dart';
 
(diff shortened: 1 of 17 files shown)


mobile-app/lib/services/reversible_transfer_monitoring_service.dart
```diff
@@ -44,14 +53,16 @@ class ReversibleTransferMonitoringService {
         .where((tx) => tx.status == ReversibleTransferStatus.SCHEDULED)
         .toList();
 
+    print(
```
dewabisma, line 56: is it still used? otherwise we should remove it

mobile-app/lib/services/reversible_transfer_monitoring_service.dart
dewabisma, line 1: Mostly the print statement, is it still used?

dewabisma (approved):
Looks good to me! Just some confirmation about the the print statement and the timer optional access that is avoiding error on null.

n13:
Lets remove print statements when we have more confidence it's always working. I need them at the moment. Not using debugPrint because that doesn't work.

PR #167, opened 2025-08-06
Reversible Max Update

n13:
When user toggles reversible, we just change the max amount  (effective balance)

Reversible transactions require there to be an existential deposit 

Screens below

Difference to before
- We don't add existential deposit to the fee (it's not a fee)
- We show the correct balance while the transaction is in flight (broadcast)


n13:
Simulator Screenshot - iPhone 16e - 2025-08-06 at 21 41 20
Simulator Screenshot - iPhone 16e - 2025-08-06 at 21 42 07
Simulator Screenshot - iPhone 16e - 2025-08-06 at 21 44 56

PR #171, opened 2025-08-07
Copy check phrase icon

### LINKED ISSUE #169: add copy button for check phrase on send screen.

n13:
Fixes #169


dewabisma (changes requested):
It should use the extension that we have.

Another thing to think about is whether to use InkWell instead of GestureDetector. From what I found out, GestureDetector using the image or rendered object alpha instead of the widget size to detect interaction. So, InkWell will be more easier to tap if that is what we wanted.

dewabisma (approved):
LGTM!

PR #181, opened 2025-08-15
fix account discovery

### LINKED ISSUE #180: Import bug
Steps:
- Import new account
- Tap on the accounts symbol going to accounts list

Expected result
- Accounts with balances appear in accounts list

Actual result
- Accounts list is empty

n13:
Fixes #180


dewabisma (approved):
Looks good! Just need to fix the analyzer error

PR #183, opened 2025-08-20
Startup fix

n13:
- Fix bug where when user creates new wallet, they get a "reload" error screen
- Fix bugs when logging out - there were widget errors due to missing themes (iPad refactor)


dewabisma (approved):
LGTM!

PR #187, opened 2025-09-04
feat: finish updating app logo in main page, navbar, and send overlay

dewabisma:
I have updated the logo in the app. I couldn't do more than this for now since we can't export anything from the figma.

image


n13 (approved):
LGTM

PR #192, opened 2025-09-09
Update onboarding screens

dewabisma:
I have updated the onboarding screen to follow new design. The only missing screen will be the setting screen after the wallet created. 

I also created few widgets that we can use to make the app feels consistent. It's encouraged to start extracting reusable widget so we don't have multiple implementation for the same UI.

Below screenshots of the new onboarding:

- Welcome screen
image

- Import wallet screen
image

- Create wallet screen
image

- Recovery phrase sheet
image


n13 (approved):
LGTM just one comment, maybe we can move at least some of these magic numbers into constants (50)?

n13:
PS I merged main with new miner app code that passes analysis

dewabisma:
> LGTM just one comment, maybe we can move at least some of these magic numbers into constants (50)?

All right, I should have done that. It's repeating anyway.

PR #199, opened 2025-09-17
Update to new branding

dewabisma:
Updated styling to match new design in figma. There are too many screens to screenshot for this PR. It's all the rest of the screens. Better to build and see directly.

I also updated the splash screen + change ticker to QU!


n13 (approved):
Nice!

PR #200, opened 2025-09-17
Heisenberg chain version

n13:


n13:
schroedinger > heisenberg

PR #206, opened 2025-09-22
Use safe area so we don't interfere with screen controls

### LINKED ISSUE #205: Android Buttons issue - Use SafeArea for main widget?
User feedback:

"I tested the Android application and noticed that the three navigation buttons at the bottom overlap with the app interface."

This was feedback in the Telegram. I think most users have those nav buttons disabled in favour of gesture control - so its good to know this.

Posting here so it doesn't slip through the net.

I think we can generate Android Simulators that show this behavior, hopefully it's just a matter of wrapping everything in SafeArea. 

SafeArea figures out device specific screen boundaries, camera hole / notch, bottom space needed for device UX, and so on.

n13:
Fixes #205

- Added safe area
- Made nav bar smaller since we no longer need to fudge the bottom offset

Some screenshots with Android with and without nav buttons, and iOS

# Android with nav buttons
Screenshot_1758553583

# Android without nav buttons
Screenshot_1758553627

# Android setting for nav buttons
Screenshot_1758553608

# iOS
Simulator Screenshot - iPhone 16e - 2025-09-22 at 23 14 51


dewabisma (approved):
Okay looking good!👍🏽 

Just need to make sure it works well for action sheet also.

PR #207, opened 2025-09-23
High security account UI + Mock integration

dewabisma:
I have finished the basic UI and mock integration. There are still many sharp edges like error handling and others.

Below is the screenshots.

- Account Setting
image

- Get Started
image

- Guardian Setup
Simulator Screenshot - iPhone 16 - 2025-09-23 at 19 54 52

- Safeguard Window Setup
Simulator Screenshot - iPhone 16 - 2025-09-23 at 19 55 13

- Recovery Setup
Simulator Screenshot - iPhone 16 - 2025-09-23 at 19 55 34

- Summary 
Simulator Screenshot - iPhone 16 - 2025-09-23 at 19 55 44

- Confirmation
Simulator Screenshot - iPhone 16 - 2025-09-23 at 19 55 53

- Success
Simulator Screenshot - iPhone 16 - 2025-09-23 at 19 56 24

PR #209, opened 2025-09-24
Feat/high security account

dewabisma:
Continuation of this PR #207 . We decided to move it away from main branch for now because we wanted to have clear separation for the new features development so we can bug fix safely in main without adding more variable into it unnecessarily.

This branch already implemented the feedback asked by removing recovery from the high security setup flow.


quantus_sdk/lib/src/models/high_security_form.dart
```diff
@@ -0,0 +1,19 @@
+class HighSecurityForm {
```
n13, line 1: What is this? Should probably have a better name if it is a model...

n13 (approved):
One comment - rename high security form... can me high security form model?

PR #218, opened 2025-09-26
Bugfix/restore account migration

n13:
- made many things async
- fixed logout which also had error


n13:
> seems okay, except that Step1:, Step 2, ai comment

It was late at night 😅

Thanks, I removed those. Really bad comments.

Issue #223, opened 2025-09-29
iPhone 8 issues (small phone issues)


n13:
Some minor layout problems - run on iPhone 8 simulator

![Image](https://github.com/user-attachments/assets/eec75c73-7373-429f-849a-fd489463ace1)

![Image](https://github.com/user-attachments/assets/b1f4e7b4-5caf-441b-9840-002bc0286acd)

No problems here

![Image](https://github.com/user-attachments/assets/dfaba6a6-75ec-40d9-bd68-d845c90d494f)

PR #225, opened 2025-09-30
fix: overflow screens

dewabisma:
Related issue #223 

I changed some font size, padding, and spacing. Also adding scroll wrapper for screen need keyboard interaction. Couldn't test in iPhone 8 but I tested iPhone 12 mini has same width only different height. But should be fine I checked SS for height overflow only 23px, I already subtract up to 70px in theory should be fine ✌️ 

Screenshots:
image

image

### DIFF
--- a/mobile-app/lib/features/components/card_info.dart
+++ b/mobile-app/lib/features/components/card_info.dart
@@ -32,11 +32,9 @@ class CardInfo extends StatelessWidget {
           if (label != null) ...[Label(label!), const SizedBox(height: 4)],
           Container(
             width: double.infinity,
-            padding: EdgeInsets.only(
-              top: verticalPadding,
-              left: 10,
-              right: 18,
-              bottom: verticalPadding,
+            padding: EdgeInsets.symmetric(
+              vertical: verticalPadding,
+              horizontal: 10,
             ),
             decoration: BoxDecoration(color: context.themeColors.buttonGlass),
             child: Row(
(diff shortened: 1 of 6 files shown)


mobile-app/lib/features/main/screens/show_recovery_phrase_screen.dart
```diff
@@ -54,7 +54,10 @@ class _ShowRecoveryPhraseScreenState extends State<ShowRecoveryPhraseScreen> {
                   const SizedBox(height: 18),
                   _buildCopyToClipboard(),
                   const SizedBox(height: 30),
-                  if (!_isRevealed) _buildWarning(),
+                  if (!_isRevealed) ...[
+                    _buildWarning(),
+                    const SizedBox(height: 24),
```
n13, line 59: why size box? is this so screen doesn't jump when revealing the phrase?

mobile-app/lib/features/components/card_info.dart
```diff
@@ -48,7 +46,7 @@ class CardInfo extends StatelessWidget {
                   width: context.isTablet ? 660 : 251,
                   child: Text(
                     text,
-                    style: context.themeText.smallParagraph?.copyWith(
+                    style: context.themeText.tag?.copyWith(
                       color: textColor,
```
n13, line 50: is this in the screenshot? did you make the text even smaller?

mobile-app/lib/features/main/screens/show_recovery_phrase_screen.dart
dewabisma: To put some space between the button and the scroll end otherwise they are tighly close almost overlap

PR #236, opened 2025-10-15
JWT login

n13:
Refactoring to put all access points into task master service

### DIFF
--- a/quantus_sdk/lib/src/services/taskmaster_service.dart
+++ b/quantus_sdk/lib/src/services/taskmaster_service.dart
@@ -103,9 +103,9 @@ class TaskmasterService {
 
   final SettingsService _settings = SettingsService();
   final HdWalletService _hd = HdWalletService();
-  String? _sessionKey;
-  String? get accessToken => _sessionKey;
-  bool get isLoggedIn => _sessionKey != null;
+  String? _accessToken;
+  String? get accessToken => _accessToken;
+  bool get isLoggedIn => _accessToken != null;
 
   TaskMasterAuthClient get _client =>
       TaskMasterAuthClient(AppConstants.taskMasterEndpoint);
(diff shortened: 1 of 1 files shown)


mobile-app/lib/services/referral_service.dart
```diff
@@ -111,46 +105,11 @@ class ReferralService {
...
-      },
-      body: jsonEncode(requestBody),
-    );
-
-    if (response.statusCode != 200) {
-      throw Exception(
-        'Referral http request failed with status: ${response.statusCode}. Body: ${response.body}',
-      );
-    }
+    await _taskmasterService.submitReferral(referralCode, activeAccount);
```
dewabisma: Just realized, I don't think we should use activeAccount but account index 0 instead for this right. We only want to use the index 0 for the referral program.

quantus_sdk/lib/src/services/taskmaster_service.dart
```diff
@@ -80,10 +82,21 @@ class TaskMasterAuthClient {
       signatureHex: sigHex,
     );
   }
+
+  Map<String, String> getAuthHeaders(String? sessionKey) {
```
dewabisma: Can we change the name of `sessionKey` everywhere to `token` or `accessToken`?

dewabisma (changes requested):
I don't think active account is reliable enough to make it idempotent. Let's make sure it's index 0 account.

quantus_sdk/lib/src/services/taskmaster_service.dart
```diff
@@ -80,10 +82,21 @@ class TaskMasterAuthClient {
...
+    return {'authorization': 'Bearer $sessionKey'};
```
dewabisma: Nice!

quantus_sdk/lib/src/services/taskmaster_service.dart
```diff
@@ -143,11 +153,56 @@ class TaskmasterService {
     }
     try {
       _sessionKey = await loginWithAccount1();
```
dewabisma: Same here

quantus_sdk/lib/src/services/taskmaster_service.dart
```diff
@@ -80,25 +82,33 @@ class TaskMasterAuthClient {
...
+    '${AppConstants.taskMasterEndpoint}/addresses',
+  );
+
   static final TaskmasterService _instance = TaskmasterService._internal();
   factory TaskmasterService() => _instance;
   TaskmasterService._internal();
 
   final SettingsService _settings = SettingsService();
   final HdWalletService _hd = HdWalletService();
   String? _sessionKey;
```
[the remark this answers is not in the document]
n13: fixed

n13:
Changed the rest of session key instances, merging

PR #240, opened 2025-10-15
Remove old opt in sheet

n13:


dewabisma (approved):
Actually we still need it, the flow remain the same, we just display video now

PR #246, opened 2025-10-16
Continuation of the previous PR

dewabisma:
I found minor bug + make optimization


n13 (approved):
LGTM

n13 (changes requested):
Actually... the cache does not look good?

mobile-app/lib/services/referral_service.dart
```diff
@@ -87,11 +95,16 @@ class ReferralService {
   }
 
   Future<bool> getRewardProgramParticiation() async {
+    if (_rewardProgramParticipationCache != null) {
```
n13: Why do we need this cache?

PR #259, opened 2025-10-17
loop last video

n13:
loop reliably on last video

Fixes this issue:

I noticed when leaving the last video looping for a while, the loop would eventually stop after a few minutes and be a black screen and the progress bar on top would glitch. 
Simulator Screenshot - iPhone 16 Pro Max - 2025-10-17 at 13 00 38


dewabisma (approved):
🔁

PR #261, opened 2025-10-17
Make navbar even greater again

dewabisma:
image


n13 (approved):
Make navbar great again!

PR #265, opened 2025-10-22
Feature/add connectivity detector

n13:
- Connectivity connector
- Connectivity banner
- Don't show errors on balance on silent update, or any update that fails. Just stick with current balance display.
- Future PRs can make this closer to the full design in Figma - the dead cat, and so on but this will cover most really bad cases where we're disconnected and show big red error text all over the screen. 

IMG_9893

### DIFF
--- a/mobile-app/lib/providers/wallet_providers.dart
+++ b/mobile-app/lib/providers/wallet_providers.dart
@@ -44,6 +44,9 @@ final balanceProviderRaw = Provider<AsyncValue<BigInt>>((ref) {
   );
 });
 
+// Store for cached balance to return on error
+BigInt _cachedBalance = BigInt.zero;
+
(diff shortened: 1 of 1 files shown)


mobile-app/lib/features/main/screens/wallet_main/wallet_main.dart
```diff
@@ -65,6 +66,7 @@ class _WalletMainState extends ConsumerState<WalletMain> {
         }
         return ScaffoldBase(
           dim: 0,
+          padding: EdgeInsets.zero,
```
dewabisma: Rather than doing this, I think we should have it become part of the scaffold base internal implementation. Because we want to have consistent passing across all screen most of the time.

We will also need to display error in other screen that required it anyway.

PR #266, opened 2025-10-23
Update account share deeplink

dewabisma:
## What's Changed
- Add new format for deep link while persisting backward compatibility
Previously we only accept `/account/:id` but now we also accept `/account?id=<id>`. It's needed because we want to be able to show fallback showing download page if user doesn't have app or failed the app domain association.

- Add referral check data caching
I noticed on slow network (IDK my phone network is snail), it's really annoying to keep loading referral when we don't have to. So, now we cache the check so it will be instant when we click referral button in setting screen after we fetch it once.

### DIFF
--- a/mobile-app/lib/services/referral_service.dart
+++ b/mobile-app/lib/services/referral_service.dart
@@ -16,7 +16,8 @@ class ReferralService {
   final TaskmasterService _taskmasterService = TaskmasterService();
 
   bool? _rewardProgramParticipationCache;
-  String? _referralDataCache = AppConstants.referralCheckIdleCode; // Indicate not fetching anything yet as __blank__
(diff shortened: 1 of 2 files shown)


dewabisma:
Im not caching the checksum i cache the get referral code request. The slow im talking about is the network request

mobile-app/lib/services/referral_service.dart
```diff
@@ -80,18 +81,25 @@ class ReferralService {
       print('getReferralData response: ${response.body}');
 
       if (response.statusCode != 200) {
+        _referralDataCache = null;
```
n13: here we cache and return null from hereon out, no recovery.

PR #271, opened 2025-10-28
Add support for Windows

Ge0:
I am addressing two issues here.

- The 1st one: for some reason, I was unable to build the `miner-app` repo because of some path error that I have addressed in the `CMakeLists.txt` file. I am not satisfied with my patch, so if you have a better solution, I am all ears.
- The 2nd one is that the app downloaded the node binary for Linux, as Windows wasn’t supported yet.

Of course, I haven’t tested whether my modifications made some regressions for Linux/Mac users.

Consider providing me with pointers on how to contribute efficiently to this repo.

Thank you for the time you will put into reviewing my work.

### DIFF
--- a/miner-app/lib/src/services/binary_manager.dart
+++ b/miner-app/lib/src/services/binary_manager.dart
@@ -32,12 +32,12 @@ class BinaryManager {
 
   static Future<String> getNodeBinaryFilePath() async {
     final cacheDir = await _getCacheDir();
-    return p.join(cacheDir.path, _binary);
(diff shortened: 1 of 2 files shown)


Ge0:
@n13 Thank you for your review.

There are other changes to come anyway, as the miner still do not work "as is" in this current state, on Windows.

Do you want me to open another PR or shall I continue with this one?

n13:
Please merge this one and open another when ready

PR #272, opened 2025-10-29
Some more UX improvement

dewabisma:
## What's Changed
- Make loading into shimmering skeleton loading for txs
- Fix overflow in several screens on small height phone 
- Make dropdown into select action sheet for better native experience
- Fix not centered avatar gradient in account setting

## Screenshots
- Skeleton Loader Main Screen
image

- Skeleton Loader Tx Screen
Simulator Screenshot - iPhone 8 - 2025-10-29 at 16 36 36

- Select action sheet
Simulator Screenshot - iPhone 8 - 2025-10-29 at 16 36 48

### DIFF
--- a/mobile-app/lib/features/components/account_copy_action_sheet.dart
+++ b/mobile-app/lib/features/components/account_copy_action_sheet.dart
@@ -89,7 +89,6 @@ class _AccountCopyActionSheetState extends State<AccountCopyActionSheet> {
           InkWell(
             onTap: _isLoading ? null : _copyChecksum,
             child: Container(
-              width: double.infinity,
               padding: const EdgeInsets.symmetric(vertical: 16),
               child: Text(
                 _isLoading ? 'Loading...' : 'Copy Checkphrase',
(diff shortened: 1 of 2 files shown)


mobile-app/lib/features/main/screens/accounts_screen.dart
```diff
@@ -63,7 +64,7 @@ class _AccountsScreenState extends ConsumerState<AccountsScreen> {
           child: Sphere(variant: 7, size: 240.681),
         ),
       ],
-      appBar: 'Your Accounts',
+      appBar:  const WalletAppBar(title:  'Your Accounts'),
```
n13: What is the benefit of the WalletAppBar class here?

n13 (commented):
Wallet app bar comments - see there - I don't think this component makes sense as it is

This is more of a question mark: Select action sheet has a lot of code.. did you also refactor the one in wallet main to use the new component? 

Rest looks good!

n13 (approved):
Nice, let's goooo!

PR #295, opened 2025-11-13
Miner app runs on dirac

illuzen:
fixed the stats too, now it uses rpc and prometheus instead of parsing logs


dewabisma (approved):
LGTM!

Issue #300, opened 2025-11-13
Shouldn't allow empty string or just whitespace as account name


dewabisma:
![Image](https://github.com/user-attachments/assets/96ddb315-c3ff-41a9-b1db-3d0bed9c29f1)

Issue #299, opened 2025-11-13
Account name update not reflecting immediately in homepage


dewabisma:
## Description:
When updating account name if the updated account is the active account, then in homepage even if not clicking the active account to go back it should show updated account name instead of stale one

## Steps to reproduce:
Update account name

## Expected result:
Account name updated

## Actual result:
Not updated still old name

## Screenshots

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/d5bbbb93-bfd7-483b-afe3-73280092ca55" />

![Image](https://github.com/user-attachments/assets/1c7badf1-bb4c-4c6f-bc13-37aaaf6b9887)

PR #307, opened 2025-11-14
Account name input

dewabisma:
Resolving #300 #299 

- Make account name input more robust with error checking.
- Fix bug of updating active account name not reflected in home screen

## Screenshots
- Showing error on bad account name
image


n13 (approved):
Nice one!

Issue #306, opened 2025-11-14
Inifnite loop for auth interval immediately


dewabisma:
## Description:
Enabling auth that prompt for biometric auth immediately, now the auth keep prompting to auth in a infinite loop

## Steps to reproduce:
Enable biometric auth, set to immediately, then close app, open app, make it succes

## Expected result:
Screen unlocked and not prompted for auth anymore unless logging out

## Actual result:
Screen unlocked and keep prompting in a loop

## Screenshot

![Image](https://github.com/user-attachments/assets/9bb42d41-61f7-473b-93b4-241279952133)

Issue #303, opened 2025-11-13
Auth always route to homepage in protected screen


dewabisma:
## Description:
Enabling auth that prompt for biometric auth in certain interval, when in a protected screen like send screen and the auth timeout is hit, it asked for biometric auth, after the auth is success it will route user to homepage.

## Steps to reproduce:
Enable biometric auth, set to 1 minute, then go to send screen, wait for 1 minute and asked for auth prompt, make it succes

## Expected result:
Screen unlocked in the same screen (send screen)

## Actual result:
Screen unlocked but it routed us to homepage

PR #308, opened 2025-11-14
Fix inifnite loop on immediately auth timeout + always send to home screen on success auth

dewabisma:
1. #303  - Resolved auth wrapper always send to home screen on success. Previously because the wrapper contain wallet initializer, it will show the home screen on success instead of the current active route. Fixed this by making auth wrapper a dumb widget.
2. #306   - Resolved infinite auth prompt on auth timeout set to immediately. The issue is on the behavior of app cycle check always happen and the invisible issue of stacked auth wrapper with each other, for example from home screen to send screen will make two auth wrapper in the widget tree result in unexpected behavior. Now, make auth wrapper dumb and maintain auth state as global provider.


n13 (approved):
Tested a bit, works

Code is cleaner than before.

LGTM!

Issue #314, opened 2025-11-20
Shouldn't be allowed to reverse tx that has reached 0 timer


dewabisma:
## Description:
Timer in-app not reliable enough to say tx is completed when sending reversible transfer even if timer is 0 by quickly reverse the transfer it's still possible.

## Steps to reproduce:
Send reverisble transfer, wait until timer 0, tap reverse

## Expected result:
Button reverse disabled or prevent reversing

## Actual result:
Still able to reverse tx if done quickly

Issue #313, opened 2025-11-20
Decimal not working without typing 0 at the beginning


dewabisma:
## Description:
When just moving to send screen and want to directly send fractional amount of token, it's not possible to just tap the decimal button, we have to type the 0 first.

## Steps to reproduce:
Route to send screen, and type decimal

## Expected result:
The amount will display `0.`

## Actual result:
Nothing happened

Issue #283, opened 2025-11-05
Implement better error display for failed transaction


dewabisma:
## Description:
When we got a failed transaction error, we need to display the error so user can report or ask about it to tech support. This error display need to be good and not annoying to see.

## Steps to reproduce:
Make a failed transaction

## Expected result:
A beautiful error display about what happened to the request

## Actual result:
Just print to console without displaying to user

Issue #281, opened 2025-11-04
Implement transfer retry button logic


dewabisma:
## Description:
When clicking retry button in transaction failed overlay nothing happen.

# Steps to reproduce:
Make a failed transaction

## Expected result:
Initializing new transfer with the same details

## Actual result:
Nothing happened

PR #316, opened 2025-11-20
Some UX improvements + Retry feature implementation

dewabisma:
## Related Issues
#314 #313 #283 #281 

## Summary
- Add generic snackbar for warning, success, error
- Add error display widgets
- Make reverse button disabled on timer zero
- Implement retry logic for failed tx
- Handle decimal tap on send automatically prepend 0 if tapping on empty value 0

## Screenshots
- Failed tx with display error button
Simulator Screenshot - iPhone 8 - 2025-11-20 at 15 27 14

- Error details sheet
Simulator Screenshot - iPhone 8 - 2025-11-20 at 15 30 02

### DIFF
--- a/mobile-app/lib/features/components/error_details_display.dart
+++ b/mobile-app/lib/features/components/error_details_display.dart
@@ -4,7 +4,6 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:quantus_sdk/quantus_sdk.dart';
 import 'package:resonance_network_wallet/features/components/button.dart';
-import 'package:resonance_network_wallet/features/components/sphere.dart';
 import 'package:resonance_network_wallet/features/styles/app_colors_theme.dart';
 import 'package:resonance_network_wallet/features/styles/app_text_theme.dart';
 import 'package:resonance_network_wallet/shared/extensions/media_query_data_extension.dart';


mobile-app/lib/features/components/transaction_action_sheet.dart
```diff
@@ -46,6 +46,8 @@ class _ReversibleTransactionActionSheetState extends ConsumerState<ReversibleTra
     return HumanReadableChecksumService().getHumanReadableName(address);
   }
 
+  bool get _isReverseDisabled => _remainingTime == null || _remainingTime == Duration.zero;
+
```
n13: is there a legitimate case where _remainingTime == null? That seems like an error?

n13: If there's a legit reason, no problem. 

If it's an error, then that variable should be non-nullable instead.

mobile-app/lib/features/components/transaction_details_action_sheet.dart
```diff
@@ -95,6 +102,41 @@ class _TransactionDetailsActionSheetState
     return 'received from';
   }
 
+  void _retryFailedTx() async {
```
n13: I guess this is referring to a transaction that failed to schedule... ?! in case of reversible that is...

quantus_sdk/lib/src/extensions/decimal_input_filter.dart
```diff
@@ -4,15 +4,17 @@ class DecimalInputFilter extends TextInputFormatter {
...
     // Allow empty input
     if (newValue.text.isEmpty) {
       return newValue;
     }
 
+    // Allow single commas or dots
+    if (newValue.text == ',' || newValue.text == '.') {
+      return TextEditingValue(text: '0${newValue.text}');
+    }
+
```
n13: you tested this yes?

Also comma and dot are treated the same yes?

mobile-app/lib/features/components/transaction_action_sheet.dart
```diff
@@ -93,7 +95,7 @@ class _ReversibleTransactionActionSheetState extends ConsumerState<ReversibleTra
             ),
           ),
           Container(
-            height: MediaQuery.of(context).size.height * AppConstants.sendingSheetHeightFraction,
+            height: MediaQuery.of(context).size.height * 0.83,
```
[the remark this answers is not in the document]
dewabisma: in iphone 8 short phone, overflowed when showing reverse confirmation.

quantus_sdk/lib/src/extensions/decimal_input_filter.dart
dewabisma: Yes, I tested it, they both treated the same as before, this line just allow user to directly tap comma or dot without having to type 0 first to create fractional value

mobile-app/lib/features/components/transaction_details_action_sheet.dart
n13: I am saying we don't get information if we manage to schedule a transaction but it then fails when it's trying to execute.

n13 (approved):
Ok looks good. Should try to make this variable non-nullable though, if there's no known reason it is nullable. That's the entire point of nullability - preventing null pointer exceptions ;)

Issue #175, opened 2025-08-10
Ensure minimum amount send


n13:
sending less than existential deposits to an account with 0 balance will always fail

don't allow in the send UX

We should probably also have failed transaction in the UX

n13:
Maybe we just hardcode this edge case for now - I have a feeling it won't affect any real users who're not testing.

PR #320, opened 2025-11-23
Improve Send Screen maintainability + Safe guard send amount

dewabisma:
Resolving #175 

## What's Changed
- Prevent sending below existential deposit, 
- Refactor send to be more readable and testable, 
- Add unit test and widget test for send screen

## Note
I have already making sure test pass and also tested it using simulator again to make sure it works and no regression.


mobile-app/lib/features/main/screens/send/send_screen.dart
```diff
@@ -242,7 +236,8 @@ class SendScreenState extends ConsumerState<SendScreen> {
       return;
     }
 
-    final parsedAmount = _formattingService.parseAmount(value);
+    final formattingService = ref.read(numberFormattingServiceProvider);
```
n13: What is the point of making all these providers?

These services provide static functions, they don't change, they're not dynamic?

dewabisma: This is part of the refactor to make the screen testable, by using provider instead of directly calling the service, we can then override it using a Mock because now it's a DI not tightly coupled.

n13: I see. Good enough reason for me! Very cool.

n13 (approved):
LGTM

PR #333, opened 2025-12-02
Add X association

dewabisma:
## Summary
Now user can associate and remove X account to their quan address.

## Screenshots
- X account not linked
Simulator Screenshot - iPhone 8 - 2025-12-02 at 12 01 13

- Associated X
Simulator Screenshot - iPhone 8 - 2025-12-02 at 12 01 05

- Removal Confirmation
Simulator Screenshot - iPhone 8 - 2025-12-02 at 12 01 08

- Quests Screen
Simulator Screenshot - iPhone 8 - 2025-12-02 at 12 01 01

### DIFF
--- a/mobile-app/lib/features/main/screens/account_associations_screen.dart
+++ b/mobile-app/lib/features/main/screens/account_associations_screen.dart
@@ -89,8 +89,8 @@ class _AccountAssociationsScreenState extends ConsumerState<AccountAssociationsS
         _isStartingOauth = true;
       });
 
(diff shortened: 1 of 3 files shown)


n13 (changes requested):
Some issues
 
- I don't think its a good design the use jwt token for auth and x auth - see other PR for potential solution but not sure that will work

It's two very different things and should exist in different storage spaces. 

Its also tedious to check purpose its not a good separation of concerns

- return results should be json for consistency 

- timeout of 5s too aggressive 

See other PR but didn't really have time to test this so not sure it will work ...

PR #337, opened 2025-12-15
[1/2] Multiple wallet support

n13:
Services updates to allow multiple wallets
Add keystone hardware wallet as an account type

### DIFF
--- a/quantus_sdk/lib/src/services/settings_service.dart
+++ b/quantus_sdk/lib/src/services/settings_service.dart
@@ -179,15 +179,15 @@ class SettingsService {
     return accounts.isEmpty;
   }
 
-  String getMnemonicKey(int mnemonicIndex) => mnemonicIndex == 0 ? 'mnemonic' : 'mnemonic_$mnemonicIndex';
+  String getMnemonicKey(int walletIndex) => walletIndex == 0 ? 'mnemonic' : 'mnemonic_$walletIndex';
 
   // Mnemonic Settings - Using secure storage
-  Future<void> setMnemonic(String mnemonic, int mnemonicIndex) async {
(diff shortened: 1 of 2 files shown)


quantus_sdk/lib/src/services/account_discovery_service.dart
```diff
@@ -21,7 +21,9 @@ class AccountDiscoveryService {
 
     // Add raw account
     final rawKeyPair = _substrateService.nonHDdilithiumKeypairFromMnemonic(mnemonic);
+    final baseWalletIndex = 0;
```
dewabisma: So, this discover account is only for loading the local account? which also mean walletIndex 0 is always local account not keystone or other wallet

quantus_sdk/lib/src/services/settings_service.dart
```diff
@@ -179,13 +179,15 @@ class SettingsService {
...
   // Mnemonic Settings - Using secure storage
-  Future<void> setMnemonic(String mnemonic) async {
-    await _secureStorage.write(key: 'mnemonic', value: mnemonic);
+  Future<void> setMnemonic(String mnemonic, int mnemonicIndex) async {
+    await _secureStorage.write(key: getMnemonicKey(mnemonicIndex), value: mnemonic);
   }
 
-  Future<String?> getMnemonic() async {
-    return await _secureStorage.read(key: 'mnemonic');
+  Future<String?> getMnemonic(int mnemonicIndex) async {
```
dewabisma: mnemonicIndex is wallet index right?

n13:
I don't think it's incorrect, it should create a new account index for each account, so the Gemini thing is wrong

PR #341, opened 2025-12-17
Fix pending transaction action sheet text

n13:
2 bugs fixed

- Fix pending transaction action sheet text
- fix timer crash when dismissing action sheet


dewabisma (approved):
Great!

Issue #343, opened 2025-12-17
Funky Auth Timer


n13:
## Description:

Auth timer is funky

When we lock the app, we need to start the timer at that point. 

What we currently do is, the timer starts from the last successful auth. 

## Steps:
- Set auth timer to 1 minute
- Use the app for 50 seconds
- Background the app for 20 seconds
- Come back to the app

## Expected: 
No authentication required, less than 1 minute away

## Actual: 
Auth screen pops up

## Reason: 
It's counting from the last successful authentication time, and not resetting this timer when we lock the app. 

I tried to fix it but there's an issue it interferes with the immediate auth workaround we had to prevent infinite looping when auth is set to immediate. Maybe we should just make immediate 100ms and call it a day? 

## Infinite auth loop
This fixed infinite auth:
#308 

Infinite auth in the case of auth timeout 0 was caused by the fact that the app is backgrounded by the FaceID overlay - so the auth user interface shown by the operating system backgrounds and foregrounds the app. 

This is why we had an infinite loop of authentication happening. 

We need to also keep this in mind.

PR #349, opened 2026-01-06
Fix auth timer

dewabisma:
## Related Issue
#343 

## Key Changes

**1. Authentication Logic Shift (`LocalAuthService`)**
-   **Old Behavior:** Checked if the time since the *last successful authentication* exceeded the timeout.
-   **New Behavior:** Checks if the time since the *app was paused (backgrounded)* exceeds the timeout.
-   **Implementation:**
    -   Added `updateLastPausedTime()` (called when app locks) and `_cleanLastPausedTime()` (called after successful auth).
    -   `shouldRequireAuthentication()` now calculates the difference between `DateTime.now()` and `getLastPausedTime()`. If `lastPausedTime` is null, it assumes no auth is needed (e.g., fresh start or already authenticated session).

**2. Lifecycle State Management (`AppLifecycleManager`)**
-   Added `_isBackgrounded` boolean to debounce lifecycle state changes. This prevents duplicate actions when the OS transitions through multiple background states (e.g., `inactive` -> `hidden` -> `paused`).
-   **On Background:** Now explicitly calls `localAuthNotifier.lockApp()` which triggers the timestamp update.
-   **On Resume:** Calls `localAuthNotifier.checkAuthentication()` which evaluates the new time-difference logic.

**3. State Provider Updates (`LocalAuthProvider`)**
-   `lockApp()`: Now records the pause time via `_localAuthService.updateLastPausedTime()` and sets `isAuthenticated` to `false`.
-   `checkAuthentication()`: explicitly sets `isAuthenticated` to `false` before prompting for biometrics, likely to ensure the UI reflects the locked state immediately.

**4. Storage (`SettingsService`)**
-   Added persistence for `last_paused_time` in `SharedPreferences` to survive app restarts if the OS kills the app while in background.

## Summary
This change fixes the issue where users might have been prompted to re-authenticate prematurely based on when they *started* their session, rather than how long they were away. It implements a true "inactivity/background timeout."

### DIFF
--- a/mobile-app/lib/app_lifecycle_manager.dart
+++ b/mobile-app/lib/app_lifecycle_manager.dart
@@ -19,10 +19,18 @@ class AppLifecycleManager extends ConsumerStatefulWidget {
 }
 
 class _AppLifecycleManagerState extends ConsumerState<AppLifecycleManager> with WidgetsBindingObserver {
+  // Track if we have already performed background/pause actions to avoid duplicates
+  // as the OS can cycle through multiple states (inactive -> hidden -> paused)
+  bool _isBackgrounded = false;
+
   @override
   void initState() {
     super.initState();
     WidgetsBinding.instance.addObserver(this);
+
+    // Initialize background state
+    final currentState = WidgetsBinding.instance.lifecycleState ?? AppLifecycleState.resumed;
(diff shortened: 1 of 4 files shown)


n13 (approved):
Code looks fine

Issue #326, opened 2025-11-26
Glitchy Updates when reversible is sent


n13:
The way we update a reversible entering into history is glitchy and looks bad on the UX. 

Do it differently.

Right now the old transaction might go away before the new one is there (pending removed, but history not updated yet)
Or it might stick around too long and you see it double. 

```
await _ref.read(pendingTransactionReconciliationServiceProvider        // Remove after a short delay to show completion // glitching here
        Timer(const Duration(seconds: 2), () {
          _ref.read(pendingTransactionsProvider.notifier).remove(pendingTx.id);
        });
```

PR #358, opened 2026-01-08
History glitch fix 

n13:
Fixes: #326 

## Cause
We deleted the pending tx with a 2 second delay, then refreshed history

## Fix
Now we explicitly update history lists with the new transaction we already found, then cause the update.

As a result, the pending transaction turns into a permanent transaction immediately, and there's no time where it suddenly disappears, or appears twice.


n13:
Github drama, closing

dewabisma (commented):
LGTM

PR #360, opened 2026-01-08
More intelligent transaction history update

### LINKED ISSUE #326: Glitchy Updates when reversible is sent

n13:
Fixes #326

## Cause
We deleted the pending tx with a 2 second delay, then refreshed history

## Fix
Now we explicitly update history lists with the new transaction we already found, then cause the update.

As a result, the pending transaction turns into a permanent transaction immediately, and there's no time where it suddenly disappears, or appears twice.


dewabisma (approved):
Nice, more better UX!

PR #362, opened 2026-01-09
High Sec Account flow

n13:
Moved from develop branch


dewabisma (approved):
Personally LGTM! 

I don't think there is a problem even in Gemini review, just some recommendations

PR #382, opened 2026-02-13
Add remote notifications

dewabisma:
## Summary
I have finished adding FCM and permission needed in iOS. I have tested that it works in iOS but I haven't tested for android. To test it, we need real device. Later we will just need to handle the kind of message and behavior we want, all the ground works are done.

## Changes
* Add necessary permission to iOS
* Add firebase configurations
* Add FCM service
* Add senoti service to register device

### DIFF
--- a/mobile-app/lib/providers/wallet_providers.dart
+++ b/mobile-app/lib/providers/wallet_providers.dart
@@ -136,3 +136,19 @@ final highSecurityEstimatedFeeProvider = FutureProvider.family<BigInt, Account>(
   );
   return feeData.fee;
 });
+
+final isBalanceHiddenProvider = StateNotifierProvider<IsBalanceHiddenNotifier, bool>((ref) {
+  final settingsService = ref.watch(settingsServiceProvider);
+  return IsBalanceHiddenNotifier(settingsService);
+});
+
+class IsBalanceHiddenNotifier extends StateNotifier<bool> {
+  final SettingsService _settingsService;
+
+  IsBalanceHiddenNotifier(this._settingsService) : super(_settingsService.isBalanceHidden());
+
+  Future<void> setIsBalanceHidden(bool value) async {
+    await _settingsService.setBalanceHidden(value);
+    state = value;
+  }
+}
--- a/mobile-app/lib/v2/screens/activity/activity_screen.dart
+++ b/mobile-app/lib/v2/screens/activity/activity_screen.dart
@@ -1,8 +1,10 @@
+import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:quantus_sdk/quantus_sdk.dart';
 import 'package:resonance_network_wallet/providers/account_providers.dart';
 import 'package:resonance_network_wallet/providers/active_account_transactions_provider.dart';
+import 'package:resonance_network_wallet/providers/wallet_providers.dart';
 import 'package:resonance_network_wallet/services/transaction_service.dart';
 import 'package:resonance_network_wallet/v2/components/back_button.dart';
 import 'package:resonance_network_wallet/v2/components/gradient_background.dart';
@@ -20,6 +22,7 @@ class ActivityScreen extends ConsumerWidget {
     final text = context.themeText;
     final accountAsync = ref.watch(activeAccountProvider);
     final txAsync = ref.watch(activeAccountTransactionsProvider);
+    final isBalanceHidden = ref.watch(isBalanceHiddenProvider);
 
     return Scaffold(
       backgroundColor: colors.background,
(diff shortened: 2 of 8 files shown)


n13:
print is fine but the crash might be an issue

n13 (commented):
Great, I now have the full picture with all 12 commits. Let me compile the updated review.

---

## PR #382 Review -- Third Pass (Latest)

The PR has gone through significant iteration since my last two reviews. Let me go through what's been addressed and what remains.

---

### Previously Raised Issues -- Now Fixed

1. **DRY: Notification tap navigation** -- `navigateToTransactionFromPayloadIfPossible` is now extracted to `TransactionService` and called from all three locations. Clean.

2. **`jsonDecode` crash on FCM tap** -- Fixed. The method now takes `Map<String, dynamic>?` directly. Local notification callers decode the String first, FCM callers pass `message.data` directly. No more type mismatch.

3. **Unused `Ref ref` parameter** -- Removed. The method uses the class's own `_ref`.

4. **`push`/`remote` cases unified** -- Clean fall-through:

```dart
case NotificationSource.push:
case NotificationSource.remote:
  _localNotificationsService.showOrScheduleNotification(notification);
  break;
```

5. **Dead `addRemoteNotification` stub removed** -- Gone, along with the stale "placeholder for future" comment.

6. **Dead `getMainAccount()` removed** from `SenotiService`.

7. **Firebase files added to `.gitignore`** -- Both `**/lib/firebase_options.dart` and `**/ios/Runner/GoogleService-Info.plist` are excluded.

8. **Feature flag** -- Everything is gated behind `FeatureFlags.enableRemoteNotifications` (currently `false`) in `main.dart`, `app_initializer.dart`, and `app.dart`. This is a great addition -- it means the feature is safely dormant until ready.

---

### Remaining Items

#### 1. Senoti endpoint is still `localhost`

```dart
static const String senotiEndpoint = 'http://localhost:3100/api';
```

Since the feature flag is `false`, this won't be hit at runtime right now. But when the flag is flipped, this will fail on real devices. Needs a production URL before enabling.

**Severity:** Low (blocked by feature flag), but a blocker for turning the flag on.

#### 2. `print()` statements in `senoti_service.dart`

Five raw `print()` calls remain, some logging sensitive data:

```dart
print('request challenge');
print('challenge: $ch');
print('msg: $msg');                    // logs the full signature message
print('verify ${ch['temp_session_id']!} $senotiEndpointUrl');
print('verify response: ${r.body}');
```

These will show in release builds (unlike `debugPrint`). The signature message and session IDs are sensitive. Again behind the feature flag, but should be cleaned up before enabling.

**Severity:** Low (behind flag), but a security concern when enabled.

#### 3. No error handling on `registerDevice` in token refresh listener

In `_setupTokenRefreshListener()`:

```dart
_messaging.onTokenRefresh.listen((newToken) {
  _senotiService.registerDevice(newToken, Platform.operatingSystem);
});
```

If senoti is unreachable, this throws an unhandled async error from a stream listener, which could crash the app. The `init()` call is wrapped in `AppInitializer`'s try-catch so `_getToken()` failures are caught, but the refresh listener has no such protection.

**Suggestion:** Wrap in try-catch:

```dart
_messaging.onTokenRefresh.listen((newToken) async {
  try {
    await _senotiService.registerDevice(newToken, Platform.operatingSystem);
  } catch (e) {
    debugPrint('Failed to re-register device: $e');
  }
});
```

**Severity:** Medium -- this can crash in production when the flag is on.

#### 4. Missing Android `google-services.json`

The `firebase.json` references `android/app/google-services.json` but this file isn't in the PR. Android builds will fail when the flag is turned on and Firebase tries to initialize. iOS has `GoogleService-Info.plist` but Android is missing its equivalent.

**Severity:** Low (feature is off + not tested on Android per the PR description), but needs to be added before Android testing.

#### 5. Info.plist whitespace churn

Still a full rewrite of every line (tabs to spaces). The actual content additions are just `UIBackgroundModes` and `FirebaseAppDelegateProxyEnabled`. Cosmetic but makes the diff noisy.

**Severity:** Cosmetic only.

---

### Summary

| Status | Item |
|---|---|
| **Fixed** | DRY: tap navigation extracted to `TransactionService` |
| **Fixed** | `jsonDecode` crash on FCM Map payload |
| **Fixed** | Unused `Ref ref` parameter removed |
| **Fixed** | `push`/`remote` cases unified |
| **Fixed** | Dead `addRemoteNotification` stub removed |
| **Fixed** | Dead `getMainAccount()` removed |
| **Fixed** | Firebase files gitignored |
| **New (good)** | Feature flag gates all FCM code (currently off) |
| **Remaining** | Senoti endpoint is `localhost` (blocker for flag-on) |
| **Remaining** | `print()` with sensitive data (security concern for flag-on) |
| **Remaining** | Unhandled error in token refresh listener (potential crash) |
| **Remaining** | Missing Android `google-services.json` |
| **Remaining** | Info.plist whitespace churn (cosmetic) |

### Verdict

The feature flag makes this safe to merge as-is -- nothing will execute at runtime with `enableRemoteNotifications = false`. All the blocking bugs from previous reviews are fixed. The remaining items (#1-#4) must be addressed before the flag gets flipped to `true`, but they don't block this PR from landing. I'd approve this with a note that items 1-3 need resolution before enabling the feature.

PR #385, opened 2026-02-16
Toaster!

dewabisma:
## Summary
Create toaster component and its' helpers. Also added extensions for easy usage.

## Screenshots
Simulator Screenshot - iPhone 8 - 2026-02-16 at 11 27 47


mobile-app/lib/v2/components/glass_container.dart
```diff
@@ -49,7 +53,12 @@ class GlassContainer extends StatelessWidget {
...
             if (filled)
               Positioned.fill(
-                child: DecoratedBox(decoration: BoxDecoration(color: Colors.white.withValues(alpha: 0.1), borderRadius: BorderRadius.circular(14))),
+                child: DecoratedBox(
+                  decoration: BoxDecoration(
+                    color: Colors.white.withValues(alpha: 0.1),
+                    borderRadius: BorderRadius.circular(14),
+                  ),
+                ),
               ),
```
n13: I don't understand these format changes - we have a common format? 

Use melos format?

n13:
Here's my review of **PR #385: Toaster!**

---

## Overview

This PR introduces a new toaster notification system to replace the existing snackbar pattern in v2 screens. It adds a `Toaster` widget, helper functions, `BuildContext` extensions, and migrates several screens from `snackbar_helper` to the new toaster. The architecture is solid overall -- nice clean layered approach with widget, helper, and extensions.

---

## Issues

### 1. DRY Violation: `toaster_helper.dart` duplicates `snackbar_helper.dart`

The `showToaster` function is nearly identical to `showTopSnackBar` -- same `flash` package, same `FlashBar` configuration with all the same transparent colors, `FlashPosition.top`, `Clip.none`, etc. The only difference is the content widget. Consider extracting the shared `FlashBar` scaffolding into a common function that accepts a `Widget content` parameter, and having both helpers call it.

### 2. Incomplete Migration

A few spots were missed:

- **`wallet_ready_screen.dart` line 94** still calls `showCopySnackbar` and still imports `snackbar_helper.dart`:

```94:94:mobile-app/lib/v2/screens/create/wallet_ready_screen.dart
      if (mounted) showCopySnackbar(context, title: 'Error', message: 'Error saving wallet: $e');
```

- **`settings_screen.dart`** still imports and uses `snackbar_helper.dart` / `showTopSnackBar`.
- **`clipboard_extensions.dart`** still retains the old `ClipboardExtensions on Clipboard` class with `copyTextWithSnackbar` and its `snackbar_helper.dart` import. If the v2 screens are fully migrated, this old extension (and the import) could be cleaned up in this PR or flagged as tech debt.

### 3. Inconsistent Extension Coverage

`toaster_extensions.dart` wraps `showSuccessToaster`, `showWarningToaster`, and `showErrorToaster` -- but **not** `showCopyToaster` or the base `showToaster`. Meanwhile, `copyTextWithToaster` lives in a separate extension (`ClipboardWithToasterExtensions`) in a different file. This makes the API surface fragmented. Consider either:
- Adding `showCopyToaster` to `ToasterExtensions`, or
- Having `copyTextWithToaster` delegate through the extension rather than importing `toaster_helper.dart` directly.

### 4. `Toaster` Widget -- Using `Icon` as a Parameter Type

```6:8:mobile-app/lib/v2/components/toaster.dart
class Toaster extends StatelessWidget {
  final String message;
  final Icon? icon;
```

Accepting `Icon?` and then destructuring it (`icon?.icon`, `icon?.color`) is a bit unusual. Cleaner alternatives:
- Accept `IconData?` and `Color?` as separate params, or
- Accept `Widget? icon` and render it directly (more flexible, allows custom widgets).

### 5. Questionable Default Icon

```14:16:mobile-app/lib/v2/components/toaster.dart
    final Widget displayIcon = Icon(
      icon?.icon ?? Icons.copy,
      color: icon?.color ?? Colors.white,
```

Defaulting to `Icons.copy` when no icon is provided means every toaster without an explicit icon shows a copy icon. An `Icons.info_outline` or no icon at all would be a safer default.

### 6. Leftover Comment

```21:21:mobile-app/lib/v2/components/toaster.dart
      // width: 343, // Width will be handled by the flash package's constraints
```

This commented-out code should be removed.

### 7. `persistent: true` -- Intentional?

```17:18:mobile-app/lib/v2/components/toaster_helper.dart
    duration: duration,
    persistent: true,
```

In the `flash` package, `persistent: true` means the flash bar won't dismiss on background tap. Users typically expect to dismiss toasters by tapping. Was this intentional?

### 8. Misleading Parameter Name

```13:13:mobile-app/lib/v2/components/toaster_helper.dart
  FlashBehavior style = FlashBehavior.floating,
```

The parameter `style` maps to `behavior` in `FlashBar`. Renaming to `behavior` would be more self-documenting and consistent with the library's API.

### 9. Redundant Default

```31:31:mobile-app/lib/v2/components/toaster.dart
        mainAxisAlignment: MainAxisAlignment.start,
```

`MainAxisAlignment.start` is the default for `Row` -- can be removed.

---

## Nits (Non-blocking)

- **Unrelated changes in this PR**: The `glass_container.dart` asset rename (`glass_button_40_bg.png` -> `glass_40.png`) and formatting-only changes in `app.dart`, `transaction_service.dart`, and `activity_section.dart` are noise. Ideally these would be in a separate cleanup PR.
- **Border radius**: `14.22` is a very specific number. If this came from a Figma spec, that's fine -- just making sure it's intentional and not a rounding artifact.

---

## Summary

The toaster component itself looks good -- clean widget, nice use of the theme system for colors, and the migration of call sites is heading in the right direction. The main concerns are:

1. **DRY**: Extract the shared `FlashBar` scaffolding between `toaster_helper` and `snackbar_helper`
2. **Completeness**: Missed a couple of migration spots in v2 screens
3. **API consistency**: The extension coverage is uneven across the helper functions
4. **`Icon` parameter pattern**: Should be `Widget?` or `IconData?` + `Color?`

Happy to help fix any of these if you'd like!

mobile-app/lib/features/main/screens/app.dart
```diff
@@ -38,7 +38,7 @@ class _ResonanceWalletAppState extends ConsumerState<ResonanceWalletApp> {
       if (FeatureFlags.enableRemoteNotifications) {
         ref.read(firebaseMessagingServiceProvider).setupNotificationTapHandlers(navigatorKey);
       }
-      
+
```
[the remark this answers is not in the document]
n13: was created by melos format

mobile-app/lib/v2/components/glass_container.dart
n13: v3 branch is off - resolved

n13:
1. and 2. should be fixed

PR #387, opened 2026-02-16
Add firebase options

dewabisma:
## Summary
Add firebase options file. We can set the api key value using .env like below.
```
IOS_FIREBASE_API_KEY=api-key
ANDROID_FIREBASE_API_KEY=api-key
```


n13 (approved):
Nice

PR #389, opened 2026-02-18
feat: add github action to release APK artifact

dewabisma:
## Summary
I added github action to release APK artifact as downloadable. I have also added secrets variables needed in the workflow to the repo.

### DIFF
--- a/.github/workflows/create_wallet_android_build.yml
+++ b/.github/workflows/create_wallet_android_build.yml
@@ -44,6 +44,7 @@ jobs:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
+            quantus_sdk/native/target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: |
             ${{ runner.os }}-cargo-


n13:
Here's my review of **PR #389: feat: add github action to release APK artifact**.

---

## Overview

This PR adds a new GitHub Actions workflow (`create_wallet_android_build.yml`) for building the Android APK and optionally creating a GitHub release. It's well-structured and follows the patterns from the existing `create_miner_build.yml`. Nice work overall -- here are the things I'd flag:

---

## Issues

### 1. Secret injection directly in shell commands (Security)

Secrets are interpolated directly into `run:` commands via `${{ secrets.* }}`. If a command fails, the expanded value can leak into logs. GitHub masks known secret values, but it's still best practice to pass them through environment variables:

```yaml
# Current (risky)
- name: Create .env file
  run: echo "${{ secrets.MOBILE_APP_ENV }}" > mobile-app/.env

# Recommended
- name: Create .env file
  env:
    MOBILE_APP_ENV: ${{ secrets.MOBILE_APP_ENV }}
  run: echo "$MOBILE_APP_ENV" > mobile-app/.env
```

Same applies to `GOOGLE_SERVICES_JSON`, `ANDROID_KEYSTORE_BASE64`, and the key.properties secrets.

**Note:** The existing `create_miner_build.yml` handles secrets via `env:` for notarization -- would be good to follow that pattern here too.

### 2. No Rust/Cargo caching

The Rust toolchain is set up with 4 Android targets (`aarch64`, `armv7`, `x86_64`, `i686`), and `quantus_sdk` uses `flutter_rust_bridge` + `cargokit` which compiles Rust during the Flutter build. Without caching `~/.cargo` and the Cargo target directory, this will be **very slow** on every run. Consider adding:

```yaml
- name: Cache Cargo
  uses: actions/cache@v4
  with:
    path: |
      ~/.cargo/registry
      ~/.cargo/git
    key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
    restore-keys: ${{ runner.os }}-cargo-
```

### 3. Flutter version not pinned

```yaml
- name: Set up Flutter
  uses: subosito/flutter-action@v2
  with:
    channel: "stable"
    cache: true
```

This uses whatever the latest stable Flutter version is at build time. Builds could break or behave differently when Flutter releases a new stable. Consider pinning a specific version (`flutter-version: "3.x.x"`) for reproducibility. The miner workflow has the same issue, so this is a repo-wide concern.

### 4. Deprecated release actions

`actions/create-release@v1` and `actions/upload-release-asset@v1` are both **archived/deprecated**. They still work, but won't receive updates. The modern replacement is `softprops/action-gh-release` which combines both steps into one:

```yaml
- name: Create Release
  uses: softprops/action-gh-release@v2
  with:
    tag_name: ${{ steps.get_version.outputs.tag }}
    name: Quantus Wallet ${{ steps.get_version.outputs.version }}
    files: ./artifacts/quantus_wallet_*.apk
```

**Note:** The miner workflow uses the same deprecated actions, so this is an existing pattern. Could be a follow-up improvement for both workflows.

### 5. DRY: Duplicate version reading

The "Read version from pubspec.yaml" step is duplicated across `build-android` and `create-release` jobs. You could pass the version as a job output from `build-android`:

```yaml
build-android:
  outputs:
    version: ${{ steps.get_version.outputs.version }}
    tag: ${{ steps.get_version.outputs.tag }}
```

Then reference it in the release job as `${{ needs.build-android.outputs.version }}`, eliminating the second checkout + grep.

### 6. `contents: write` permission is global

The `contents: write` permission is set at the workflow level, but only the `create-release` job needs write access (for tagging and creating releases). The `build-android` job only needs `contents: read`. Consider scoping permissions per-job for least-privilege:

```yaml
jobs:
  build-android:
    permissions:
      contents: read
    ...
  create-release:
    permissions:
      contents: write
    ...
```

---

## Minor Nits

- The `find ./artifacts -type f | sort` step is purely for debugging -- consider removing it or guarding it behind a debug flag.
- The release body says "See commit history for details" which isn't super helpful. Maybe auto-generate from commits since last tag, or at least link to the compare view.

---

## What looks good

- Clean `workflow_dispatch` with choice inputs for build mode and release toggle
- Smart conditional: releases only allowed in `release` mode (`inputs.create_release && inputs.build_mode == 'release'`)
- Tag collision detection before pushing
- APK renaming with version and mode in the filename
- Proper use of `melos bootstrap` for the monorepo (vs the miner workflow's direct `flutter pub get`)

---

**Verdict:** Solid workflow that follows existing project patterns. The secret handling (#1) and missing Rust caching (#2) are the most impactful items to address before merge. The rest can be follow-ups.

n13 (approved):
LGTM

PR #390, opened 2026-02-18
Finalize remote notification

dewabisma:
## Summary
Update the remote message handling to properly mapped into notification data known in the mobile app. Also updated the endpoint to production one.

### DIFF
--- a/mobile-app/lib/services/firebase_messaging_service.dart
+++ b/mobile-app/lib/services/firebase_messaging_service.dart
@@ -107,7 +107,6 @@ class FirebaseMessagingService {
       await _senotiService.unregisterDevice(token, _platform);
     } catch (e) {
       debugPrint('Failed to unregister device: $e');
-      rethrow;
     }
   }
 
--- a/mobile-app/lib/v2/screens/settings/settings_screen.dart
+++ b/mobile-app/lib/v2/screens/settings/settings_screen.dart
@@ -65,11 +65,7 @@ class _SettingsScreenV2State extends ConsumerState<SettingsScreenV2> {
 
   Future<void> _resetAndClearData() async {
     if (FeatureFlags.enableRemoteNotifications) {
-      try {
-        ref.read(firebaseMessagingServiceProvider).unregisterDevice();
-      } catch (e) {
-        debugPrint('Failed to unregister device (non-fatal): $e');
-      }
+      ref.read(firebaseMessagingServiceProvider).unregisterDevice();
     }
 
     _settingsService.clearAll();


n13:
The PR looks solid overall. It correctly implements the remote notification handling, updates the backend endpoint to production, and handles device registration/unregistration lifecycles.

However, I have a few observations and questions:

### 1. `pubspec.lock` Downgrades
The `pubspec.lock` file shows a downgrade in the Flutter and Dart SDK constraints:
```diff
-  dart: ">=3.10.0 <4.0.0"
-  flutter: ">=3.38.0"
+  dart: ">=3.9.0 <4.0.0"
+  flutter: ">=3.35.0"
```
Was this intentional? It looks like it might have been generated with an older version of Flutter.

### 2. `SenotiService` Account Assumption
In `quantus_sdk/lib/src/services/senoti_service.dart`, the `_getAccount1Credentials` method relies on `_settingsService.getMnemonic(0)`:

```dart:quantus_sdk/lib/src/services/senoti_service.dart
    final mnemonic = await _settingsService.getMnemonic(0);
    if (mnemonic == null) {
      throw Exception('Mnemonic not found.');
    }
```

This assumes the user always has a mnemonic-based wallet at index 0. If the app supports importing accounts via JSON or private key (where no mnemonic exists), or if the primary account isn't at index 0, this will throw an exception and registration will fail.

### 3. Error Handling in `insertNewAddress`
In `FirebaseMessagingService`, `insertNewAddress` catches errors and debug prints them:

```dart:mobile-app/lib/services/firebase_messaging_service.dart
    } catch (e) {
      debugPrint('Failed to insert new address: $e');
    }
```
This is likely fine for a background/fire-and-forget task, but just noting that failures here will be silent to the user.

### 4. Production Endpoint
You've updated `senotiEndpoint` to `https://snt.quantus.com/api`. Just confirming this is the intended production endpoint and we are ready to point to it.

### Summary
The logic changes in `FirebaseMessagingService` to use `TransactionService` for parsing are excellent and much cleaner. The lifecycle management (register/unregister) is also a great addition. If the SDK version change is fine and the mnemonic assumption holds true for the current app architecture, this is good to go.

dewabisma:
I have removed auth from registering remote notifications

n13 (changes requested):
Reset should not wait definitely not fail. 

The rest also are good

n13:
The PR looks solid and addresses the critical feedback from the previous review. Here is my analysis of the changes:

### Key Improvements & Fixes
1.  **Blocking Reset Fixed**: In `SettingsScreenV2` (`_resetAndClearData`), the `unregisterDevice` call is now wrapped in a `try-catch` block. This ensures that if the network call fails (e.g., offline or server down), the user can still reset their wallet and wipe local data.
2.  **Auth Removed**: `SenotiService` has been simplified to remove the authentication requirement for registering devices, as requested.
3.  **Dependencies**: `pubspec.lock` has been updated. The Flutter SDK constraint is now `>=3.38.0`, which is an upgrade rather than the accidental downgrade seen previously.
4.  **iOS Configuration**: The changes to `Info.plist` (migrating to `UIScene`, setting `FirebaseAppDelegateProxyEnabled` to `false`) and `AppDelegate.swift` are correct for modern Flutter Firebase Messaging setups that require background fetch and custom handling.

### Minor Suggestion (Nit)
In `mobile-app/lib/v2/screens/home/accounts_sheet.dart`:
```dart
792:       await ref.read(firebaseMessagingServiceProvider).insertNewAddress(accountToSave.accountId);
```
The `insertNewAddress` method in `FirebaseMessagingService` already catches and swallows all exceptions (printing them to debug console). Awaiting it here forces the UI to wait for the network request before closing the "Create Account" sheet, making the app feel slower.

Since the UI doesn't handle errors from this call (they are swallowed inside the service), you can safely run this unawaited to keep the UI snappy:

```dart
// Recommended: Run unawaited so the sheet closes immediately
ref.read(firebaseMessagingServiceProvider).insertNewAddress(accountToSave.accountId);
```

### Conclusion
**Approve**. The critical blockers are resolved. The minor suggestion about `insertNewAddress` can be addressed in a follow-up or before merge if desired, but it's not a blocker.

n13 (approved):
I am OK with this but please check all uses of the new token functions and make sure none of them block the UI

The UI does not have to wait for this, it's useless to do so

We have in the past often managed to add unnecessary wait loops without user interface, so let's avoid this in the future.

n13 (approved):

PR #393, opened 2026-02-18
Fix workflow wrong key.properties location

dewabisma:
Previously put the key.properties wrong and also the signer file.


dewabisma:
> Yup!

WOW you are blazing fast and always watching!

PR #396, opened 2026-02-19
Swap screen fixes

n13:
- icons for swap tokens
- quan xchange rate $1
- formatting large numbers better, so we can show 0.0000012 BTC for example
- no checking for refund address validity... it's just a demo

Simulator Screenshot - iPhone 16e - 2026-02-19 at 11 38 38
Simulator Screenshot - iPhone 16e - 2026-02-19 at 11 39 30

### DIFF
--- a/mobile-app/lib/v2/screens/swap/swap_screen.dart
+++ b/mobile-app/lib/v2/screens/swap/swap_screen.dart
@@ -38,7 +38,15 @@ class _SwapScreenState extends State<SwapScreen> {
   String get _rateLabel {
     final val = 1 / _rate;
     if (val == 0) return '1 QUAN = 0 ${_fromToken.symbol}';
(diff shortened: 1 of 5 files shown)


mobile-app/lib/v2/screens/swap/swap_screen.dart
```diff
@@ -449,34 +455,87 @@ class _SwapScreenState extends State<SwapScreen> {
   }
 }
 
-class _QrScanPage extends StatelessWidget {
+class _QrScanPage extends StatefulWidget {
```
dewabisma: I think in QR Scan there is also another button to get QR from gallery?

n13:
> Looking good! Just some questions.
> 
> Also, does it mean near intents properly set up already? or just for getting the rate and swap quote? Seems getting USD price still mock though

Everything is mock

We can't really use near on testnet ;)

PR #398, opened 2026-02-20
activity section added

n13:
will update this with final design from sejal

Simulator Screenshot - iPhone 16e - 2026-02-20 at 13 06 09


dewabisma (approved):
LGTM! Just some suggestions

PR #406, opened 2026-02-26
android edge to edge fix

n13:
## PR Summary

This PR fixes Android edge-to-edge and lint issues for release builds.  
It removes deprecated Flutter system bar color calls, keeps edge-to-edge enabled, and updates Android theme/resource config to satisfy `lintRelease` without changing cross-platform build portability.

### Included changes
- Removed explicit `SystemUiOverlayStyle` status/navigation bar color settings in `mobile-app/lib/main.dart`.
- Added optional camera hardware declaration (`required=false`) in `mobile-app/android/app/src/main/AndroidManifest.xml` to resolve ChromeOS-related lint error.
- Updated launch themes in:
  - `mobile-app/android/app/src/main/res/values/styles.xml`
  - `mobile-app/android/app/src/main/res/values-night/styles.xml`
  using `tools:targetApi` on API-gated attributes (`forceDarkAllowed`, `windowLayoutInDisplayCutoutMode`) so behavior is preserved and lint passes.
- Removed machine-specific Gradle JDK pin from `mobile-app/android/gradle.properties` to avoid environment-specific breakage.

### Verification
- `:app:lintRelease` passes when run with JDK 21.


dewabisma (commented):
Looking fine if it works. But a little disturbed by the different targetApi. We defined 27 and 29. Shouldn’t we stick to one?

dewabisma (approved):
All right, let's do this!

PR #407, opened 2026-03-02
Wormhole Implementation

n13:


mobile-app/pubspec.lock
```diff
@@ -909,10 +909,10 @@ packages:
     dependency: transitive
     description:
       name: image_picker_ios
-      sha256: b9c4a438a9ff4f60808c9cf0039b93a42bb6c2211ef6ebb647394b2b3fa84588
+      sha256: "956c16a42c0c708f914021666ffcd8265dde36e673c9fa68c81f7d085d9774ad"
       url: "https://pub.dev"
     source: hosted
-    version: "0.8.13+6"
+    version: "0.8.13+3"
   image_picker_linux:
```
n13: hmm why is image picker in the mobile app changed?

can we remove this change, or do we actually need this version of the picker for some reason

n13 (commented):
I think we should delete the change to the mobile lock file

Also this should be merged into the new_testnet_planck branch, not main

Main is still on old testnet (for not much longer but anyway)

Also CI failing

PR #413, opened 2026-03-03
Refactor boilerplate

dewabisma:
## Summary
I refactor quite a lot of stuffs, the main goal is to make it more opinionated in how we create new screen or sheet. This way we don't have to think much when creating them and also if we need to update spacing, padding, or background we can do it once. 

I have tested all the features and it works as it should, so there shouldn't be any regression. Also, maybe please help testing it too maybe I missed some cases. I didn't test the swap feature as it's under feature flag also.

## Changes
- adjust and move scaffold base to v2
- create v2 app bar
- refactor sheets to use standardized sheet
- update icon buttons in tx details and scanner

### DIFF
--- a/mobile-app/lib/v2/screens/accounts/accounts_sheet.dart
+++ b/mobile-app/lib/v2/screens/accounts/accounts_sheet.dart
@@ -6,7 +6,7 @@ import 'package:quantus_sdk/quantus_sdk.dart';
 import 'package:resonance_network_wallet/features/components/account_gradient_image.dart';
 import 'package:resonance_network_wallet/v2/screens/accounts/add_hardware_account_screen.dart';
 import 'package:resonance_network_wallet/services/firebase_messaging_service.dart';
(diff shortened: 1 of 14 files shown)


mobile-app/lib/v2/components/qr_scanner_page.dart
```diff
@@ -94,15 +95,7 @@ class _QrScannerPageState extends State<QrScannerPage> {
...
-    return GestureDetector(
-      onTap: onTap,
-      child: Container(
-        width: 40,
-        height: 40,
-        decoration: BoxDecoration(color: Colors.white.withValues(alpha: 0.1), borderRadius: BorderRadius.circular(8)),
-        child: Icon(icon, size: 20, color: colors.textPrimary),
-      ),
-    );
+    return ButtonIcon.rounded(icon: icon, onTap: onTap);
```
n13: renamed to GlassIconButton

PR #420, opened 2026-03-21
N13/pos v2 check pending

n13:
- Fast checking for POS mode
- Refactor to reuse code
- Clear cached mining rewards on logout


mobile-app/lib/v2/screens/activity/tx_item.dart
```diff
@@ -32,7 +32,7 @@ class TxItemData {
 
     return TxItemData(
       label: isPending
-          ? 'Sending'
+          ? (isSend ? 'Sending' : 'Receiving')
           : isScheduled
           ? (isSend ? 'Pending' : 'Receiving')
```
dewabisma, line 37: This wording become weird when we have scheduled transfer, because isPending is always the initial state.

So, by current logic the label will be,

Sending/Receiving -> Pending/Receiving

mobile-app/lib/services/pending_receive_tracker.dart
dewabisma, line 1: Is this supposed to be used by both send and swap also? I think the logic here is extracted from the tx submission logic.

dewabisma (commented):
I added comments that need to be cleared up, then it will be good to go

mobile-app/lib/v2/screens/activity/tx_item.dart
n13: we don't have scheduled for now in the current app so it's ok

it seems reasonable but once we have a design for scheduled, timer, and so on we can revisit

for now the app has no "isScheduled" state since we didn't implement reversible transactions yet

dewabisma: kind of weird to regress over it. But actually I did refactor on this nested ternary dystopia in my v3 PR, so I will fix it there.

dewabisma (approved):
LGTM!

PR #421, opened 2026-03-21
N13/pos mode with tx watch

n13:
Note: This unfortunately requires new testnet. 

I think we should make a branch of main for this. 

**Update**: Targeting new_testnet_planck - which is our new app running on planck

- POS system subscribes to transaction pool and checks incoming transactions
- Takes less than 1 second
- Block inclusion - settlement - still takes 10+ seconds and up to minutes
- This is a quick confirmation between trusted parties

pos-mode


dewabisma (changes requested):
Just need to clarify this, all good

PR #425, opened 2026-03-24
Remote feature flags

dewabisma:
## Summary
- Remove old way of constant feature flags
- Add remote feature flags provider
- Add new method to task master service to get feature flags

## Note
We shouldn't merge this before we deploy the new task master version

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,6 +33,8 @@ jobs:
         run: melos exec --concurrency=1 -- "flutter clean"
       - name: Analyze
         run: melos exec --concurrency=1 -- "flutter analyze ."
+      - name: Create mock .env file to not fail the tests
+        run: touch mobile-app/.env
       - name: Test Mobile App
         run: melos exec --concurrency=1 --scope="resonance_network_wallet" -- "flutter test"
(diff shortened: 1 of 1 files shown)


quantus_sdk/lib/src/models/feature_flags_model.dart
```diff
@@ -47,4 +53,23 @@ class FeatureFlagsModel {
 
     return value;
   }
+
```
n13: Question: Why did you not just add a separate function called "flagsHaveChanged" or whatever and use that for the same-ness comparison. 

I usually prefer doing this over overriding the == / equals / hash functions because it doesn't have any side effects, it's constrained to that one check which we want to do... it's basically a more targeted code piece that does exactly what we need it to do. This way if == is used elsewhere in the code under different circumstances we don't override that... 

Also whichever way you do it, please make a unit test that will trigger when we changed something...

ie what usually happens is, you add some flag... and then forget to update the == override, and then it's mostly working but when only that flag changes it doesn't work

There's a way to structure to the code to make sure such errors are caught at compile time or at least at testing... 

 maybe we can put all compare values in an array or something like that... not sure...

PR #426, opened 2026-03-27
renamed feature flags remote config

n13:
it's not just feature flags, that's why

Pure rename PR - nothing else changed, just class names and filenames.


quantus_sdk/test/contract/remote_config_api_test.dart
```diff
@@ -39,7 +39,7 @@ void main() {
       expect(newKeys, isEmpty, reason: 'WARNING: The API sent new properties not handled in your app: $newKeys');
 
       try {
-        FeatureFlagsModel.fromJson(data);
+        RemoteConfigModel.fromJson(data);
       } catch (e) {
         fail('Failed to parse feature flags model: $e');
```
dewabisma: also this one :)

dewabisma (approved):
LGTM, just need to update the contract test  later

PR #430, opened 2026-03-31
scan deep link QR from send page

### LINKED ISSUE #429: POS scan URL should work within the app
Steps to reproduce:
- Got o app
- Open "Send"
- Click "Scan QR" button
- Scan a POS QR code

Expected Result:
- App scans amount and destination, fills it in, shows ok button

Actual result
- App fails, the URL is put in the destination field

See video

https://github.com/user-attachments/assets/be3c76d6-a27b-4bbe-ab93-e96b37aea1db

n13:
fixes #429

- if deep link, scan and populate fields
- if valid quantus address, populate address
- if anything else - keep scanning (imagine many QR codes on a page, we don't want to fail if we pick up a bad one, we scan until we find a valid one)


dewabisma (approved):
Let's ship!

PR #435, opened 2026-04-02
check for missing mnemonic

n13:
Fix for https://github.com/Quantus-Network/quantus-apps/issues/434

- On startup, check for mnemonic!
- If missing, log user out
- Setting to store mnemonic in encrypted backups (iOS) 

Mnemonic still won't get stored in unencrypted backups, that's an iOS safety feature.

#### Testing
Change the code to always pretend like the mnemonic is missing, ie 
instead of 
`      if (mnemonic == null) {
`
do 

      if (true) {

### DIFF
--- a/mobile-app/flutter_launcher_icons.yaml
+++ b/mobile-app/flutter_launcher_icons.yaml
@@ -3,7 +3,7 @@ flutter_launcher_icons:
   image_path: "assets/icon/icon.png"
 
   android: "launcher_icon"
-  image_path_android: "assets/quantus_icon_1024.png"
+  image_path_android: "assets/./"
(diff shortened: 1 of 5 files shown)


mobile-app/flutter_launcher_icons.yaml
dewabisma: Why do we change the path? There is no changes for asset.

dewabisma (commented):
Overall looking good, just not sure that the change path about in launcher.yaml

mobile-app/flutter_launcher_icons.yaml
n13: Whoa, this is a bug, thanks! Will fix

dewabisma (approved):
LGTM!

PR #439, opened 2026-04-06
perfecting import screen

n13:
it's perfect now

- Fixed overflow
- Do the right thing when the screen is small (scroll to the import button)

Simulator Screenshot - iPhone 16 - 2026-04-06 at 14 22 18


dewabisma (approved):
LGTM!

PR #440, opened 2026-04-06
Version number

n13:
- Add version number script
- invoke script in both ios and android builds

This is better... I think.. than pulling in a 3rd party package like package_info_plus

We will see!  

Simulator Screenshot - iPhone 16 - 2026-04-06 at 15 00 21


dewabisma (approved):
LGTM

n13:
The remote test somehow fails, but passes locally - ignoring this

PR #441, opened 2026-04-08
auth improvements

n13:
Changed a few things to make auth work the way it should

We had auth flickering in for a long time, ignored on iOS but saw a very bad video of it on Android. 

So now it should all work. 

**Fix auth lock screen UX issues**

- Separate visual lock (privacy overlay) from authentication state for finer-grained control over what the user sees during app lifecycle transitions
- Require biometric auth before sending transactions
- Clear `isVisuallyLocked` on authenticate completion so the privacy overlay doesn't linger after Face ID succeeds
- Center the lock screen layout (was off-center due to `GradientBackground` Stack using loose constraints)
- Remove debug UX artifacts


dewabisma:
Can you please provide screenshots of the changes?

mobile-app/lib/services/local_auth_service.dart
```diff
@@ -45,10 +45,15 @@ class LocalAuthService {
 
       final didAuthenticate = await _localAuth.authenticate(
         localizedReason: localizedReason,
-        options: const AuthenticationOptions(biometricOnly: false, stickyAuth: true, sensitiveTransaction: true),
+        options: const AuthenticationOptions(
+          biometricOnly: false,
+          stickyAuth: true,
+          sensitiveTransaction: true,
+          // Use default useErrorDialogs: true for better native UX on Android/iOS
```
dewabisma: What is this comment for?

dewabisma (approved):
LGTM!

mobile-app/lib/services/local_auth_service.dart
n13: It's confusing, removing this ;)

PR #443, opened 2026-04-09
Restore ultra secure security

n13:
- iPhone backups or data transfers to a new phone will **never** transfer the secret phrase.

This was the default security setting before I changed it recently, restoring it.

### DIFF
--- a/mobile-app/pubspec.lock
+++ b/mobile-app/pubspec.lock
@@ -1085,10 +1085,10 @@ packages:
     dependency: transitive
     description:
       name: matcher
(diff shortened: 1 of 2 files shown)


dewabisma (commented):
SO we go back to square one? Why do we have this pubspec change again?? every PR seems to change this, can you please investigate it?

n13:
Probably new flutter version
No square one, just secure mode

dewabisma (approved):
I see, yeah I thought before it was already without the iOptions thing. Maybe I forgot. LGTM

PR #444, opened 2026-04-09
Update graphql query and data accessors

dewabisma:
## Summary
- Now extrinsic is a data type so we adjust the way we access it in transfer and reversible transfer data model
- Updated query in chain history to access the extrinsicHash correctly


n13 (approved):
LGTM!

PR #445, opened 2026-04-09
Android renderer hotfix

n13:
Android liquid glass renderer is broken

This replaces it with a fake glass renderer based on the fake renderer inside of the liquid_glass_renderer package

**Fix Android rendering artifacts in glass components**

The `liquid_glass_renderer` package uses GPU compositing layers (`BackdropFilterLayer`) that produce visual artifacts on Android -- tearing in glass buttons and icon buttons disappearing entirely during scroll.

This PR adds `AndroidGlass`, a pure-widget replacement that replicates the glass specular highlights using standard `CustomPaint` + `ClipRRect`, avoiding the problematic compositing layer. `LiquidGlassBase` now routes to `AndroidGlass` on Android while keeping the full `LiquidGlass` effect on iOS.

**Changes:**
- `liquid_glass_base.dart` -- platform gate: iOS uses `LiquidGlass`, Android uses `AndroidGlass`
- `android_glass.dart` (new) -- self-contained glass widget with specular highlight painter, no dependency on `liquid_glass_renderer`

### Magic number explanation

`0x1AFFFFFF` isn't magic -- it's just your existing `surfaceGlass` theme color from `app_colors.dart`:

```98:99:mobile-app/lib/v2/theme/app_colors.dart
        surfaceGlass: const Color(0x1AFFFFFF),
        surfaceCard: const Color(0x0FFFFFFF),
```

It's white at 10% opacity. We use it as the fallback when callers pass `Colors.transparent` (the default), since a fully transparent glass wouldn't be visible without the backdrop blur that the real `LiquidGlass` provides.

### DIFF
--- a/mobile-app/lib/services/local_auth_service.dart
+++ b/mobile-app/lib/services/local_auth_service.dart
@@ -45,11 +45,7 @@ class LocalAuthService {
 
       final didAuthenticate = await _localAuth.authenticate(
         localizedReason: localizedReason,
-        options: const AuthenticationOptions(
-          biometricOnly: false,
-          stickyAuth: true,
-          sensitiveTransaction: true,
-        ),
+        options: const AuthenticationOptions(biometricOnly: false, stickyAuth: true, sensitiveTransaction: true),
       );
 
       if (didAuthenticate) cleanLastPausedTime();
--- a/mobile-app/lib/v2/components/liquid_glass_base.dart
+++ b/mobile-app/lib/v2/components/liquid_glass_base.dart
@@ -1,3 +1,4 @@
+import 'dart:io' show Platform;
 import 'package:flutter/material.dart';
 import 'package:liquid_glass_renderer/liquid_glass_renderer.dart';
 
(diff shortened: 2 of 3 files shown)


mobile-app/lib/services/local_auth_service.dart
```diff
@@ -45,11 +45,7 @@ class LocalAuthService {
 
       final didAuthenticate = await _localAuth.authenticate(
         localizedReason: localizedReason,
-        options: const AuthenticationOptions(
-          biometricOnly: false,
-          stickyAuth: true,
-          sensitiveTransaction: true,
-        ),
+        options: const AuthenticationOptions(biometricOnly: false, stickyAuth: true, sensitiveTransaction: true),
```
dewabisma, line 48: Funny we have this formatting change

mobile-app/lib/v2/components/liquid_glass_base.dart
```diff
@@ -30,31 +32,38 @@ class LiquidGlassBase extends StatelessWidget {
...
+    if (Platform.isAndroid) {
+      return AndroidGlass(
+        visibility: settings.visibility,
+        glassColor: settings.glassColor == Colors.transparent ? const Color(0x1AFFFFFF) : glassColor,
+        radius: radius,
+        centered: centered,
+        child: child,
+      );
+    } else {
+      Widget glass = LiquidGlassLayer(
```
dewabisma: This put the order of center not the same as before, this one smell. I think there is a reason of the order of the widget before to make the rendering correct. Please make it same order.

dewabisma (changes requested):
Found one smell change, please see comment

mobile-app/lib/v2/components/liquid_glass_base.dart
n13: yeah there's no reason to make this change this confusing whether that is the same or not - fixed

mobile-app/lib/services/local_auth_service.dart
n13: IDK i just run melos run format, it did this

n13: maybe it was wrong before

dewabisma (approved):
Superb!

PR #449, opened 2026-04-14
New testnet planck

n13:
new testnet is LIVE

- [x] - change RPC access points
- [x] - change subsquid URL
- [x] - update metadata
- [x] - update rust packages
- [x] Run known value keys tests
- [x] - make sure migration works
- [x] Fix migration dialog style
- [x] Fix account pending index in reversible

## Known Value tests fix 

**Root cause:** The SDK's `to_account_id` was using `PoseidonHasher::hash` (which internally calls `hash_for_circuit` - a ZK-circuit-optimized hash with compact encoding and padding). The chain uses `hash_bytes` (injective encoding, 4 bytes/felt, no padding). These are explicitly different functions - the `qp-poseidon-core` test suite even asserts they produce different output.

**Changes made:**

1. **`rust/Cargo.toml`** - Replaced `qp-poseidon` with `qp-poseidon-core` (direct dependency on the core crate that has `hash_bytes`)

2. **`rust/src/api/crypto.rs`** - Changed `to_account_id` to use `hash_bytes` (matching the chain's `AccountId32::new(hash_bytes(pubkey))`) instead of `PoseidonHasher::hash`

3. **`test/generate_keys_test.dart`** - Uncommented the master key assertion, updated the "known values2" expected address to match the corrected output

New migration dialog

Tested error cases too

Simulator Screenshot - iPhone 16 - 2026-04-14 at 19 05 39


dewabisma (approved):
I have faith it works as expected 🤞

PR #454, opened 2026-04-17
Fix new account screen

n13:
Fix the new account screen - had a bug where name could not be changed

Simulator Screenshot - iPhone 17 - 2026-04-17 at 22 05 35


n13:
self approved

PR #456, opened 2026-04-19
Show up to date mining stats

n13:
added planck stats by directly querying the chain

Simulator Screenshot - iPhone 17 - 2026-04-19 at 14 51 55


n13:
That makes sense, so I’d drop that as a blocker.

The only remaining issue I’d keep from the review is the CI formatting failure in `quantus_sdk/lib/src/services/taskmaster_service.dart`. The branch is still failing the `dart format --set-exit-if-changed` check, so it needs a small cleanup before merge.

PR #458, opened 2026-04-22
Redesign send screen

dewabisma:
## Summary
- Send screen now becomes multi screen process
- Refactor repeating address checkphrase with avatar into reusable widget
- Many more

## Screenshots
- Select recipient screen
Simulator Screenshot - iPhone 8 - 2026-04-22 at 22 30 31

- Select recipient screen with selected
Simulator Screenshot - iPhone 8 - 2026-04-22 at 22 30 37

- Input amount screen
Simulator Screenshot - iPhone 8 - 2026-04-22 at 22 30 44

- Input amount screen with valid amount
Simulator Screenshot - iPhone 8 - 2026-04-22 at 22 30 47

- Review send screen
Simulator Screenshot - iPhone 8 - 2026-04-22 at 22 30 50

- Tx submitted screen
Simulator Screenshot - iPhone 8 - 2026-04-22 at 22 30 52


mobile-app/lib/v2/components/address_checkphrase_with_initial.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:resonance_network_wallet/v2/theme/app_colors.dart';
+import 'package:resonance_network_wallet/v2/theme/app_text_styles.dart';
+
+class AddressCheckphraseWithInitial extends StatelessWidget {
+  final String recipientChecksum;
+  final String recipientAddress;
+
+  const AddressCheckphraseWithInitial({super.key, required this.recipientChecksum, required this.recipientAddress});
+
+  String _initials(String? checksum) {
```
n13, line 12: we have checksum initials?

mobile-app/lib/v2/screens/send/send_screen_logic.dart
```diff
@@ -56,7 +56,7 @@ class SendScreenLogic {
       case AmountStatus.insufficientBalance:
         return 'Insufficient Balance';
       case AmountStatus.valid:
-        return 'Send ${formattingService.formatBalance(amount, addSymbol: true)}';
+        return 'Review Send';
```
n13, line 59: ?

n13:
I've gone through the full diff carefully. Here's my review of **PR #458 - Redesign send screen**:

---

## Overall

Great refactor! Breaking the monolithic `SendSheet` bottom sheet into a proper multi-screen flow (`SelectRecipient` -> `InputAmount` -> `ReviewSend` -> `TxSubmitted`) is a solid architecture improvement. The extraction of `AddressCheckphraseWithInitial` and `AmountDisplayWithConversion` into reusable widgets is aligned with DRY principles, and the exchange rate decimal fix (`3,96` -> `3.96`) squashes a real bug.

---

## n13's Review Comment

The comment on line 12 of `address_checkphrase_with_initial.dart` -- **"we have checksum initials?"** -- is a valid question. The old `AddressPickerSheet` used `AccountGradientImage` (a unique gradient circle derived from the account ID) as the avatar. The new `AddressCheckphraseWithInitial` replaces that with text initials extracted from the checksum (e.g. "Rainbow-Sunset" -> "RS"). This loses visual uniqueness -- any two addresses whose checksum words start with the same letters will have identical avatars. The gradient approach was more robust for distinguishing addresses at a glance. Worth considering keeping `AccountGradientImage` here, or using it as a fallback background behind the initials.

---

## Issues Found

**1. Force-unwrap crash risk in `_openReview()`**

In `InputAmountScreen`, `_recipientChecksum!` is force-unwrapped. The button is disabled when `_recipientChecksum == null`, but there's a potential race window between the async checksum lookup completing and the user tapping the button.

```dart
// input_amount_screen.dart, inside _openReview
recipientChecksum: _recipientChecksum!,
```

**2. Global `InputDecorationTheme` change could break other screens**

The `app_theme.dart` changes set `fillColor: Colors.transparent`, `contentPadding: EdgeInsets.zero`, and `isDense: true` globally. Any existing text fields in the app that relied on the previous defaults (filled surface color, default padding) will be affected.

**3. `AddressFormattingService.formatAddress` default change is a breaking SDK change**

The defaults changed from `prefix: 5, postFix: 5` to `prefix: 6, postFix: 6`. This affects every call site using the defaults across the entire codebase, including the SDK consumers.

**4. `isSend` changed from `required` to optional with default `true`**

In `txAmountDisplayProvider`, `isSend` went from `required bool` to `bool isSend = true`. This silently changes behavior: any call site that previously omitted `isSend` would get a compile error, now it silently defaults to showing a "-" prefix. Verify all call sites are intentional.

**5. `sendSectionLabel` iPad size not scaled**

Both `defaultTheme` and `iPad` constructors define `sendSectionLabel` with `fontSize: 20`. Other styles like `paragraph` scale from 16 -> 20 for iPad. Looks like an oversight.

**6. Network fee not shown on `InputAmountScreen`**

The old `SendSheet` displayed the network fee on the form step. The new `InputAmountScreen` fetches the fee but only shows it on the `ReviewSendScreen`. Users can't see the fee before proceeding -- minor UX regression.

**7. Raw `TextStyle` instead of theme in `_recipientCard`**

The "SEND TO" label in `InputAmountScreen._recipientCard()` uses a raw `TextStyle(...)` instead of a theme text style. This is inconsistent with the rest of the codebase's theming pattern and won't scale with iPad sizing.

---

## Minor Notes

- The `_amountController` in `SelectRecipientScreen` is only set from QR scan but passed to `InputAmountScreen` as `initialAmount` -- naming is slightly misleading since it's not visible on the select screen itself.
- The deep navigation stack (4 screens pushed) is cleaned up correctly via `popUntil` in `TxSubmittedScreen` -- nice.
- The `activityGroupLabel` -> `receiveLabel` swap works because both had identical style definitions. Clean.

---

n13 (commented):
Review added

mobile-app/lib/v2/components/address_checkphrase_with_initial.dart
dewabisma: Yeah in the figma design we do have, you can check the select recipient screen screenshot I attached.

mobile-app/lib/v2/screens/send/send_screen_logic.dart
dewabisma: Just changing the wording because now we have a overview screen of the tx before confirming to send

dewabisma:
No 2,3,4 is intentional.

```
1. Force-unwrap crash risk in _openReview()
```

I will handle this better, seems AI not happy with my stingy approach.

```
5. sendSectionLabel iPad size not scaled
```
Whoopsss missed it.

```
6. Network fee not shown on InputAmountScreen
```

This is how the design is, I will tell Sejal about it. I think it worths further discussion. 

```
7. Raw TextStyle instead of theme in _recipientCard
```
Whoops another miss

n13:
4 still sounds kinda bad though - why is isend  = true the default? Why not make it explicit which one to use in each situation as it was? ... why the change? Is there a particular reason?

Overall doesn't make much of a difference, it's not super important certainly not a blocker, but curious. 

I use defaults when there's an existing function and I am adding a parameter but don't want to change the existing code. 

Or when there's a logical default for the object to be in just because it's much more common, but send/receive seems 50/50....

dewabisma:
> 4 still sounds kinda bad though - why is isend = true the default? Why not make it explicit which one to use in each situation as it was? ... why the change? Is there a particular reason?
> 
> Overall doesn't make much of a difference, it's not super important certainly not a blocker, but curious.
> 
> I use defaults when there's an existing function and I am adding a parameter but don't want to change the existing code.
> 
> Or when there's a logical default for the object to be in just because it's much more common, but send/receive seems 50/50....

Okay, it seems having default not appropriate here since send and receive can be 50/50. I agree with that point.

n13 (commented):
## Code Review: Redesign Send Screen

Great refactor — breaking the monolithic bottom-sheet send flow into a proper multi-screen navigation (Select Recipient → Input Amount → Review → Tx Submitted) is a solid architectural improvement. The new `AmountDisplayWithConversion` and `AddressCheckphraseWithInitial` reusable widgets are good extractions. Fixing the exchange rate decimal separators (commas → dots) is a nice catch.

### DRY Concerns

1. **`_fetchEstimatedFee` / `_fetchFee` duplication in `InputAmountScreen`** — These two methods share ~80% identical code (call `balancesService.getBalanceTransferFee`, guard on `RegularAccount`, set `_networkFee` and `_blockHeight`, handle errors identically). Consider extracting a single `_fetchFee({BigInt? amount, String? recipient})` that defaults to the estimated params when amount/recipient aren't provided.

2. **Recipient display built 3 different ways** — `InputAmountScreen._recipientCard`, `ReviewSendScreen._heroCard`, and `TxSubmittedScreen` each build their own checksum + address display. The new `AddressCheckphraseWithInitial` widget exists but isn't used in any of these three screens. If the layouts differ slightly, consider parameterizing `AddressCheckphraseWithInitial` rather than re-building it from scratch each time.

3. **`ReviewSendScreen` shows address twice** — The hero card displays TO + address, and the summary section below also has a `_summaryRow(label: 'TO', value: shortAddr)`. The user sees the recipient address in two places stacked on top of each other.

### Architectural Notes

- **Screen-to-screen data passing via constructor args** — `InputAmountScreen` accepts 4 params, `ReviewSendScreen` accepts 6 params, all passed forward through Navigator pushes. For this linear flow, consider a shared send-flow state object (e.g. a `ChangeNotifier` or a scoped Riverpod provider) to avoid threading data through every constructor. This will become more important if the flow grows.

- **`InputAmountScreen` uses `ScaffoldBase(child: ...)` but the next PR renames this to `mainContent`** — Fine since these are stacked PRs, just noting for awareness.

- Good call keeping `SendScreenLogic` as a pure logic helper and reusing it here.

Overall a clean decomposition — the main actionable items are the fee-fetching duplication and the missed reuse of `AddressCheckphraseWithInitial`.

n13 (approved):
Approved, but please check if the above items are worth addressing - doesn't seem critical though.

n13:
image
Also this

PR #459, opened 2026-04-24
Fix migration bug

n13:
- fix migration bug - update active account provider
- remove stats poll service which we no longer use


dewabisma (approved):
LGTM!

PR #464, opened 2026-04-26
Beast/redesign home screen again

dewabisma:
## Summary
-  Update balance view
-  Update tx item

## Screenshots
- Home balance
Simulator Screenshot - iPhone 8 - 2026-04-26 at 23 40 05

- Home balance flipped
Simulator Screenshot - iPhone 8 - 2026-04-26 at 23 39 59


n13 (commented):
## Code Review: Redesign Home Screen Again

Small, focused PR. Adds the QUAN logo to the balance display and refines transaction item colors. Clean changes.

### DRY Concerns

1. **Three color-deriving methods in `TxItemData`** — `labelColor`, `amountColor`, and the existing icon color all follow the same pattern: check `isIncoming` vs `isOutgoing`, check `isHighlighted`, return a color from the theme. The branching logic is similar across all three. Consider a single `_resolveColor({required Color incomingHighlight, required Color incoming, required Color outgoingHighlight, required Color outgoing, required Color fallback})` helper to reduce the conditional duplication.

2. **`_quantusLogoImage` built inline in `AmountDisplayWithConversion`** — The `SvgPicture.asset('assets/v2/uppercase_q.svg', ...)` is currently inline. If the QUAN logo is used anywhere else (and it likely will be in a crypto app), extract a small `QuanLogo({double size})` widget or at minimum define the asset path as a constant. This avoids silent breakage if the asset gets renamed/moved.

### Architectural Notes

- **`useQuanLogo` boolean flag** — Adding a boolean to `AmountDisplayWithConversion` is fine for now, but if more display variants accumulate (different icons, different prefix/suffix styles), consider an enum or a slot-based approach instead of boolean flags.

- **Color rename `txItemOutgoingHighlight` → `txItemOutgoingHighlightBorder`** — Good semantic precision. The old name was ambiguous about whether it was for text, background, or border.

- **No state changes, no new providers** — Pure UI refinement. Keeps the blast radius small.

Looks good. Minor DRY items only.

n13 (approved):
AI review stuff is really minor, probably not worth addressing

Approved

PR #468, opened 2026-04-29
Remove total count

n13:
- remove total count
- more digits for balances by default (4 instead of 2)
- kept home screen at 2 but we will need to think about this

### DIFF
--- a/quantus_sdk/lib/src/services/chain_history_service.dart
+++ b/quantus_sdk/lib/src/services/chain_history_service.dart
@@ -5,9 +5,16 @@ import 'package:quantus_sdk/quantus_sdk.dart';
 
 class OtherTransfersResult {
   final List<TransactionEvent> transfers;
-  final int totalCount;
+  final bool hasMore;
 
-  OtherTransfersResult({required this.transfers, required this.totalCount});
+  const OtherTransfersResult({required this.transfers, required this.hasMore});
+}
+
+class _Page<T> {
+  final List<T> items;
+  final bool hasMore;
+
+  const _Page({required this.items, required this.hasMore});
 }
 
 class ChainHistoryService {
@@ -132,9 +139,6 @@ query AccountEvents($accounts: [String!]!, $limit: Int!, $offset: Int!) {
         hash
       }
     }
-  }
-  accountEventsConnection(orderBy: id_ASC, where: {AND: [{account: {id_in: $accounts}, balanceEvent_isNull: true, scheduledReversibleTransfer_isNull: true}, {OR: [{transfer_isNull: true}, {transfer: {extrinsic_isNull: false}}]}]}) {
-    totalCount
   }
 }
 ''';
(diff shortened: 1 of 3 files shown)


quantus_sdk/lib/src/services/chain_history_service.dart
```diff
@@ -491,22 +504,22 @@ query SearchByExtrinsicHash($extrinsicHash: String!) {
...
-        fetchScheduledReversibleTransfers(accountIds: accountIds, limit: limit, offset: scheduledOffset),
+        _fetchScheduledReversibleTransfersPage(accountIds: accountIds, limit: limit, offset: scheduledOffset),
         fetchOtherTransfers(accountIds: accountIds, limit: limit, offset: otherOffset),
       ]);
 
-      final scheduledReversibleTransfers = results[0] as List<ReversibleTransferEvent>;
+      final scheduledReversibleTransfers = results[0] as _Page<ReversibleTransferEvent>;
       final otherTransfers = results[1] as OtherTransfersResult;
 
-      final nextOtherOffset = otherOffset + limit;
```
dewabisma, line 501: Not really sure why we need to change this.

dewabisma (commented):
Just need to clarify one thing about changing how we calculate offset. Other than that, looks good! We don't really need the totalCount in mobile since we don't use numeric pagination but infinite scroll.

quantus_sdk/lib/src/services/chain_history_service.dart
n13: limit and result length are different things?!

maybe we set limit to 100, but got only 2 back for whatever reason - then we have to start from there.

At the end of the list we get less than limit back. But in some environments we could also get less in other situations (I think not applicable to our db though)

n13: Overall I Don't know why limit isn't just a fixed number, seems we could simplify this code a lot...but it was dynamic before too. Could do in another PR

dewabisma: Limit is fixed, the way we calculate offset should be simple. So let's say limit is 10, first fetch offset should be 0 * limit, second fetch offset should be 1 * limit, and so on.

dewabisma: Okay, let's do that in another PR.

dewabisma (approved):
LGTM, we still need to properly address how we manage pagination. We can do that in next PR. I think using cursor based will be better since we don't need to know page number in mobile app

quantus_sdk/lib/src/services/chain_history_service.dart
n13: > Limit is fixed, the way we calculate offset should be simple. So let's say limit is 10, first fetch offset should be 0 * limit, second fetch offset should be 1 * limit, and so on.

That is not true. The offset is (last_offset + number of results we got)

The number of results may be != limit

PR #469, opened 2026-04-29
Miner: wormhole claim rewards

n13:
## Summary

End-to-end wormhole claim rewards flow for the miner app — discovers unspent transfers, generates ZK proofs, and submits on-chain claims. Same architecture as the CLI; some networking/caching choices differ.

### Bugfix
_flutter: [WormholeUtxo] getUnspentBalance: 351418140770225 planck (910 unspent transfers)_

some time later...

_flutter: [WormholeUtxo] getUnspentBalance: 352578512793634 planck (913 unspent transfers)_

-> I think we're now up to date and include newly mined blocks into the rewards total - users still have to manually hit the update button but it works

### What's included

- **`WormholeUtxoService`** (SDK) — fetches transfers from Subsquid, checks nullifiers, returns unspent set. Local cache for both transfers and spent nullifiers; cache only advances to `currentHeight - reorgDepth` so reorgs never invalidate persisted entries, but callers always see transfers up to the chain head.
- **`WormholeClaimService`** (miner) — full pipeline: fetch unspent → ensure circuit binaries → fetch merkle proof + generate ZK proof per leaf (16 in parallel) → aggregate in batches of 16 → submit each batch as an unsigned extrinsic.
- **Claim rewards UI** — stepper dialog with destination input, confirmation, real-time progress, and result/cancel summary.
- **Rust FFI wormhole module** — `generate_proof`, `aggregate_proofs`, `compute_merkle_positions`, `decode_leaf_amount`, etc., bridged via flutter_rust_bridge.
- **`ChainRpcClient`** — added `getZkMerkleProof`, `getBlockHash`, `getBestBlockHash`, header helpers.

### Review fixes applied

- **Fail-loudly everywhere** — removed `5_000_000` chain-height fallback in `WormholeUtxoService`; `ChainRpcClient.{getBlockHeader,getFinalizedHead,getZkMerkleProof,getBlockHash}` throw `StateError` instead of returning `null`.
- **Cache off-by-one** — GraphQL filter is now `height_gt: cachedUpToBlock` (no `+1`), so block N+1 is no longer skipped.
- **Cache vs. display semantics** — `getTransfersTo` returns transfers up to current head; only the reorg-safe slice is persisted.
- **Proof block = current head** (not finalized), so transfers in the last `reorgDepth` blocks are claimable; reorg-induced verification failure → user retries.
- **Net total accounting** — `ClaimResult.totalWithdrawn` accumulates `outputAmount * SCALE_DOWN_FACTOR` (post-fee planck delivered on-chain).
- **Cancellation** — `IsCancelledCallback` + `WormholeOperationCancelled` plumbed through utxo service; dialog shows a distinct "Cancelled" terminal state.
- **Address validation upfront** via `getAccountId32` before the progress screen.
- **`Future.wait(eagerError: true)`** on proof chunks so the first failure aborts immediately.
- **Inner hash UI** — displays as `0x…` hex (not SS58) in settings and setup screens.
- **DRY** — consolidated hex helpers on `package:convert`, balance formatting on `NumberFormattingService`, magic numbers replaced with `_volumeFeeBps` / `_scaleDownFactor` constants.

#### 1. Mnemonic persistence is re-enabled (`MinerWalletService` + `SettingsService`)

`main` previously reverted the miner to a preimage-only flow (`03da8fd1`). This PR adds it back (`8099b668`) so the app can derive the wormhole secret and run the in-app withdraw flow without round-tripping through the CLI. Mnemonic is stored via `flutter_secure_storage`, keyed by wallet index. On macOS this requires the keychain entitlement added in this PR.

**Shipping requirement:** the macOS build must be signed with the correct Apple Developer / App Store certificate and an updated provisioning profile that includes `keychain-access-groups`. Without that, secure storage will fail at runtime on signed builds.

#### 2. TelemetryDeck SDK added (`main.dart` + `pubspec.yaml`)

Adds `telemetrydecksdk: ^3.0.0` and a `Telemetrydecksdk.start(...)` call with hardcoded `appID` and `salt`. Unrelated to claim rewards — review separately for privacy, opt-in story, and event scope.

### Known limitations

- **No hash-prefix queries** — Subsquid's `transfersByHashPrefix` / `nullifiersByPrefix` resolvers don't paginate (>1000 results errors). Workaround: `transfers(where: { to: { id_eq: ... } })` and `wormholeNullifiers(where: { nullifierHash_in: [...] })`. The Subsquid server therefore sees the exact wormhole address and nullifier hashes — privacy trade-off until prefix resolvers are fixed server-side.
- **Proof generation speed** — parallelized at 16 concurrent, but slower than the Rust CLI: Dart `Future.wait` adds event-loop scheduling overhead vs. Rayon. Heavy lifting is in Rust; overhead is purely dispatch.
- **Serial batch submission** — each aggregated batch awaits acceptance before the next. That's OK! 
- **No wait for inclusion** - waiting for block inclusion is not needed. We will anyway later find the nullifiers and know which ones have failed if any - generally transactions which have returned a transaction hash are actually verified due to our chain's pre-verification (stops DDOS of garbage proofs). Also waiting for block inclusion takes long and has almost no benefit in this case. In the worst of all cases, a proof was accepted but later doesn't get applied (chain reorg maybe, although waiting for inclusion would not fix that either) - even in this case, later on when we re-check the used nullifiers, we would find all unspent balances again.

## Test plan

- [x] Full claim flow on testnet with ~800 unspent transfers — proofs generated, aggregated, submitted (5 batches landed in destination balance)
- [x] Cancellation mid-flow stops cleanly with a distinct "Cancelled" UI state
- [ ] Zero unspent transfers (should short-circuit to "nothing to claim")
- [ ] Network interruption during proof generation
- [ ] Keychain access on a signed/notarized macOS build with updated provisioning profile

## Screens

https://github.com/user-attachments/assets/4f844f0f-2b16-40ed-a957-0b90ff99f579


ethan-crypto (commented):
### Blocking / high-priority findings

**1. Incremental transfer cache skips one block every refresh.**
`_fetchTransferPage` queries transfers with `block.height_gt: $afterBlock`, but `getTransfersTo` sets `queryFrom = cache.cachedUpToBlock + 1` before passing it as `afterBlock`. That means after caching through block `N`, the next query asks for `height > N + 1`, permanently skipping transfers in block `N + 1`. This can make rewards disappear until the user clears cache. Use either `afterBlock: cache.cachedUpToBlock` with `height_gt`, or change the GraphQL filter to `height_gte` when passing `cachedUpToBlock + 1`. Add a regression test with `cachedUpToBlock = 1000` and a transfer at `1001`. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/dcab20644f13366281a8d3cf97347bf987320bf8/quantus_sdk/lib/src/services/wormhole_utxo_service.dart)][1])

**2. RPC height fallback can poison the cache far into the future.**
`_getChainHeight` falls back to `5,000,000` on RPC failure, then `getTransfersTo` saves `cachedUpToBlock = safeCutoff`. If the real chain height is below that fallback, the cache can be advanced millions of blocks ahead; future calls will skip new transfers because `queryFrom > chainHeight`. This should fail closed: if RPC height is unavailable, do not mutate transfer cache and do not invent a safe cutoff. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/dcab20644f13366281a8d3cf97347bf987320bf8/quantus_sdk/lib/src/services/wormhole_utxo_service.dart)][1])

**3. Unsafe/recent transfers are returned even though the claim proofs are generated against the finalized head.**
`getTransfersTo` computes `safeCutoff` and only persists `safeTransfers`, but it returns `allTransfers`, including newly fetched transfers above the cutoff. The claim service then fetches a finalized block hash/header and requests Merkle proofs for every unspent transfer against that finalized block. Recent transfers that are not in the finalized tree can fail proof generation and take down the whole claim. I would return/filter only safe transfers for claiming, or expose recent transfers separately as “pending finality.” ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/dcab20644f13366281a8d3cf97347bf987320bf8/quantus_sdk/lib/src/services/wormhole_utxo_service.dart)][1])

**4. “Submitted/complete” does not mean included or finalized.**
The PR text says batches are submitted sequentially because each on-chain submission must confirm before the next one, but `_submitUnsignedProof` only calls `author_submitExtrinsic` and returns the hash; there is no inclusion/finality watch and no event/error decoding. The UI can therefore report completion after the node accepts the transaction into the pool, not after the claim succeeds on-chain. Use `author_submitAndWatchExtrinsic` or the project’s normal tx flow, wait for `InBlock`/`Finalized`, and fail on `ExtrinsicFailed` before starting the next batch. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/pull/469)][2])

### Medium-priority findings

**5. `ClaimResult.totalWithdrawn` overstates the actual destination amount.**
Each proof computes `outputAmount = wormholeComputeOutputAmount(inputAmount, feeBps: 10)` and passes that fee-adjusted amount into the proof, but the returned `ClaimResult.totalWithdrawn` is set to the sum of the raw input transfer amounts. If this result is later displayed or used for accounting, it will be wrong by the volume fee. Accumulate net output amounts instead, or rename the field to make it clear that it is pre-fee. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/dcab20644f13366281a8d3cf97347bf987320bf8/miner-app/lib/src/services/wormhole_claim_service.dart)][3])

**6. Cancellation is cooperative but not very clean during expensive work.**
The cancel flag is checked before each 16-transfer proof chunk and before each batch, but not inside the in-flight `Future.wait`, proof generation, aggregation, or submission call. A user who cancels during a proof chunk or during `_submitUnsignedProof` will still see that expensive operation finish before cancellation is observed. I would add checks after `Future.wait`, before aggregation, after aggregation, and before/after submission, and update the UI copy to avoid implying immediate cancellation. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/dcab20644f13366281a8d3cf97347bf987320bf8/miner-app/lib/src/services/wormhole_claim_service.dart)][3])

**7. Cancellation is displayed as completion.**
On `StateError` containing “cancelled,” the dialog sets `_done = true` and `_resultMessage = 'Cancelled by user'`. But the progress screen title resolves to “Claim Complete” whenever `_running == false` and `_errorMessage == null`, and `_buildStepRow` marks every step completed when `_done` is true. That will show a green complete state for a cancelled claim. Track cancellation separately, or treat it as a neutral/error terminal state. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/dcab20644f13366281a8d3cf97347bf987320bf8/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart)][4])

**8. Destination address validation is too weak.**
The dialog only rejects empty strings and addresses shorter than 40 characters. The actual conversion to account bytes happens later in the claim service, after setup and UTXO work has already started. Invalid SS58/base58 strings with sufficient length will pass the UI check and then fail later. Validate with the same SS58/account decoder before allowing the confirmation screen. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/dcab20644f13366281a8d3cf97347bf987320bf8/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart)][4])

### Lower-priority / hygiene

**9. GitHub flags hidden/bidirectional Unicode in `miner_balance_card.dart`.**
This is worth cleaning up before merge even if it is harmless formatting noise. Remove hidden characters and run the formatter so future reviewers do not have to reason about Trojan-source style ambiguity. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/pull/469/files)][5])

**10. Test coverage should be strengthened before merging.**
The PR has a manual test plan, including a full testnet claim, cancellation, zero-unspent, and network interruption cases, but the risky parts here need automated coverage: cache off-by-one, RPC-height failure not mutating cache, safe/finalized transfer filtering, net amount accounting, invalid destination validation, cancellation UI state, and submit-vs-finalized behavior. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/pull/469)][2])

### What looks good

The overall feature split is sensible: UTXO discovery/caching lives in the SDK service, claim orchestration is isolated in `WormholeClaimService`, the dialog surfaces the main phases, and batching is capped at 16 proofs consistently with the constants in the Rust API. The known privacy limitation around address/nullifier queries is also called out explicitly in the PR text, so I would treat that as an accepted product tradeoff rather than a surprise blocker. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/pull/469)][2])

[1]: https://github.com/Quantus-Network/quantus-apps/blob/dcab20644f13366281a8d3cf97347bf987320bf8/quantus_sdk/lib/src/services/wormhole_utxo_service.dart "quantus-apps/quantus_sdk/lib/src/services/wormhole_utxo_service.dart at dcab20644f13366281a8d3cf97347bf987320bf8 · Quantus-Network/quantus-apps · GitHub"
[2]: https://github.com/Quantus-Network/quantus-apps/pull/469 "Miner: wormhole claim rewards by n13 · Pull Request #469 · Quantus-Network/quantus-apps · GitHub"
[3]: https://github.com/Quantus-Network/quantus-apps/blob/dcab20644f13366281a8d3cf97347bf987320bf8/miner-app/lib/src/services/wormhole_claim_service.dart "quantus-apps/miner-app/lib/src/services/wormhole_claim_service.dart at dcab20644f13366281a8d3cf97347bf987320bf8 · Quantus-Network/quantus-apps · GitHub"
[4]: https://github.com/Quantus-Network/quantus-apps/blob/dcab20644f13366281a8d3cf97347bf987320bf8/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart "quantus-apps/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart at dcab20644f13366281a8d3cf97347bf987320bf8 · Quantus-Network/quantus-apps · GitHub"
[5]: https://github.com/Quantus-Network/quantus-apps/pull/469/files "Miner: wormhole claim rewards by n13 · Pull Request #469 · Quantus-Network/quantus-apps · GitHub"

n13:
does not work for illuzen here 
Screenshot 2026-05-04 at 12 53 08 PM

n13:
Here's my review of PR #469 — Miner: wormhole claim rewards.

I read through the actual changes (not just ethan-crypto's review) and verified each finding. Most of his comments hold up, and there are a few additional things worth flagging. Below is what I would block on, what I would push back on, and what's net-positive.

## Blocking issues (verified)

### 1. Cache off-by-one — real, will silently lose a block

```354:368:quantus_sdk/lib/src/services/wormhole_utxo_service.dart
    final queryFrom = cache.cachedUpToBlock > 0 ? cache.cachedUpToBlock + 1 : 0;
    List<WormholeTransfer> newTransfers;

    if (queryFrom > chainHeight) {
      _log('Cache is current, no new blocks to query');
      newTransfers = [];
    } else {
      _log('Querying transfers after block $queryFrom');
      newTransfers = await _fetchAllTransfers(
        toAddress: wormholeAddress,
        afterBlock: queryFrom,
```

Combined with the GraphQL filter `block: { height_gt: $afterBlock }` (line 167), if the cache is at block N, we ask for `height > N+1`, i.e. the next call only returns blocks ≥ N+2. Block N+1 is permanently skipped. Two acceptable fixes:

- pass `afterBlock: cache.cachedUpToBlock` (don't `+1`), or
- change the GraphQL operator to `height_gte` (and rename the variable).

### 2. RPC chain-height fallback poisons the cache (also violates a user rule)

```136:154:quantus_sdk/lib/src/services/wormhole_utxo_service.dart
  Future<int> _getChainHeight() async {
    try {
      final body = jsonEncode({'jsonrpc': '2.0', 'id': 1, 'method': 'chain_getHeader', 'params': []});
      final response = await _rpcEndpoint.post(body: body);
      ...
    } catch (e) {
      _log('RPC chain_getHeader failed: $e');
    }
    _log('Chain height: fallback to 5000000');
    return 5000000;
  }
```

If the RPC fails, `chainHeight = 5_000_000` → `safeCutoff = 4_999_820` → that's persisted to `cachedUpToBlock`. On every subsequent call, `queryFrom > realChainHeight` so we hit the "Cache is current" branch and never refetch. The cache is permanently frozen until the user manually wipes it.

This is exactly what your repo rule says to avoid: *"Fail early — never have a silent failure, all failures must produce error logging; never add fallback code for unexpected failures."* This needs to throw instead of fabricate a height.

### 3. Returning post-cutoff transfers will hard-fail proof generation

```376:382:quantus_sdk/lib/src/services/wormhole_utxo_service.dart
    final safeTransfers = allTransfers.where((t) => t.blockHeight <= safeCutoff).toList();
    final updatedCache = _TransferCache(cachedUpToBlock: safeCutoff, transfers: safeTransfers);
    await _saveTransferCache(fullHash, updatedCache);

    _log('getTransfersTo DONE: ${allTransfers.length} transfers (${sw.elapsedMilliseconds}ms)');
    return allTransfers;
  }
```

Only `safeTransfers` is cached, but `allTransfers` is returned. The claim service then asks for a Merkle proof against the **finalized** head:

```120:130:miner-app/lib/src/services/wormhole_claim_service.dart
    final blockHash = await rpc.getFinalizedHead();
    if (blockHash == null) throw StateError('Failed to get finalized block hash');
    final header = await rpc.getBlockHeader(blockHash: blockHash);
```

A transfer above `safeCutoff` may not yet be in the finalized tree, and `getZkMerkleProof` will return null at line 149 → `StateError` → the entire claim aborts (no batches submitted, even for the safe transfers). Either filter to `safeTransfers` for claiming, or expose recent ones separately as "pending finality".

### 4. "Submitted" is reported as "Done" even when the chain rejects the proof

```232:253:miner-app/lib/src/services/wormhole_claim_service.dart
  Future<String> _submitUnsignedProof(ChainRpcClient rpc, Uint8List aggregatedProofBytes) async {
    ...
    final result = await rpc.rpcCall('author_submitExtrinsic', [hexExtrinsic]);
    if (result == null) throw StateError('Extrinsic submission returned null');
    return result as String;
  }
```

`author_submitExtrinsic` returns the moment the node accepts the extrinsic into the pool — that's not inclusion, and definitely not success. The PR description says *"each on-chain submission must confirm before the next"* but the code doesn't actually verify any such thing. If the proof is rejected on-chain, the user sees a green check, the next batch fires, and the rewards just don't show up. Use `author_submitAndWatchExtrinsic` (or polling for `system.events::ExtrinsicFailed`) and fail the chain step on `ExtrinsicFailed`. I'd call this blocking, not medium.

## Additional issues I'd raise (not in ethan-crypto's review)

### 5. Mnemonic persistence is being re-added against a recent reversal

`miner_wallet_service.dart` now stores the mnemonic via `SettingsService.setMnemonic`. But main has these two recent commits:

```
8099b668 adding mnemonic storage back
03da8fd1 revert miner mnemonic persistence to preimage-only flow
```

The current `main` is the *revert*, and this PR puts the mnemonic back. The PR description doesn't mention it — it only mentions re-adding the keychain entitlement. If this is intentional, it should be called out in the PR title/body so reviewers know the security posture is changing again. If it isn't intentional, those changes should be peeled out.

### 6. Telemetry SDK silently added

```diff
+  Telemetrydecksdk.start(
+    const TelemetryManagerConfiguration(appID: '22C2FABE-0070-4BCA-B774-AB4380607EB2', salt: 'QDay'),
+  );
```

A new dependency `telemetrydecksdk: ^3.0.0` and a hardcoded appID/salt are introduced in `main.dart` with no mention in the PR description. For an app that just gained the ability to handle mnemonics and on-chain claims, dropping in third-party telemetry is a privacy change that needs its own review — at minimum an opt-in toggle, an explicit callout in the PR body, and probably a separate PR.

### 7. DRY violations (your repo rule)

- `_hexToBytes` is duplicated three times: `wormhole_utxo_service.dart` (line 446), `wormhole_claim_service.dart` (line 264), `chain_rpc_client.dart` (line 372).
- `_hexToInt` is in both `wormhole_claim_service.dart` (line 259) and `chain_rpc_client.dart` (line 477).
- `_formatBalance`/`_formatPlanck` are duplicated in `claim_rewards_dialog.dart` (line 51) and `miner_balance_card.dart` (line 86) with slightly different fractional widths.

The `convert` package is already a direct dependency now — `hex.decode`/`hex.encode` covers most of these. The few things that don't have a one-liner equivalent should live in a single shared util. Per your *"Strictly follow the DRY principle … duplicate code must be avoided at all costs"* rule, this should be cleaned up.

### 8. `ChainRpcClient` is full of catch-and-return-null

`getBlockHeader`, `getFinalizedHead`, `getZkMerkleProof`, `getBlockHash`, `getCurrentBlock`, `isReachable`, `isSyncing`, etc. all do `try { ... } catch (e) { return null; }`. The claim service then re-throws with a generic `StateError('Failed to get …')`, losing the original error message. Same fail-loud rule as #2 — the new methods added by this PR (`getZkMerkleProof`, `getFinalizedHead`, `getBlockHeader`, `getBlockHash`) should propagate the underlying exception so callers can show the real reason.

### 9. Cancellation isn't wired into the UTXO service at all

The cancel flag is only checked inside `_runClaimFlow`. `WormholeUtxoService.getUnspentTransfers` has no cancel hook, so cancelling during steps 2/3/4 only takes effect once it returns. With ~800 transfers, that's a lot of network work the user can't actually stop. Pass a cancellation token (or a `bool Function() isCancelled`) into the service.

### 10. `Future.wait` doesn't surface useful errors

```186:191:miner-app/lib/src/services/wormhole_claim_service.dart
          proofBytesList[i] = proof.proofBytes;
        }());
      }

      await Future.wait(futures);
```

If one proof throws (e.g. the unsafe-transfer issue from #3), the other 15 still run to completion before the chunk fails, and only the first error is surfaced. Use `Future.wait(futures, eagerError: true)` and surface the underlying error, not a `StateError`.

## Confirming ethan-crypto's findings

| Finding | Status |
|---|---|
| 1. Cache off-by-one | **Confirmed** — see #1 above. |
| 2. RPC fallback poisons cache | **Confirmed** — see #2. Also a user-rule violation. |
| 3. Returns unsafe transfers | **Confirmed** — see #3. |
| 4. Submit ≠ included | **Confirmed**, and I'd raise to blocking. |
| 5. `totalWithdrawn` is gross, not net | **Confirmed.** `outputAmount = inputAmount * (1 - feeBps/10000)` per leaf, but `totalWithdrawn = totalAmount` (sum of inputs). Either accumulate `outputAmount` or rename the field. |
| 6. Cooperative cancellation gaps | **Confirmed** — and worse than he says, because the UTXO service has no cancel at all (#9). |
| 7. Cancel shown as "Claim Complete" | **Confirmed.** Line 388-394 renders "Claim Complete" whenever `!_running && _errorMessage == null`, and line 495 marks all steps green when `_done == true`. After a cancel, all six checks turn green. Track cancel as its own terminal state. |
| 8. Weak destination validation | **Confirmed.** `length < 40` is the only check; full SS58 decode happens after circuit setup, UTXO discovery, and finalized-head fetch — multiple seconds of wasted work on a typo. Validate via `getAccountId32` (in a try/catch) before the confirm screen. |
| 9. Hidden Unicode in `miner_balance_card.dart` | **Could not reproduce.** Grepped for bidi/zero-width codepoints — file is clean ASCII. I'd treat this as a false positive from GitHub's lint, unless ethan can point to a specific line. |
| 10. Test coverage | Agree — at minimum, the cache off-by-one and the safe/unsafe transfer split should have unit tests, since they're easy to miss in a regression. |

## What's good

- Clean separation: SDK owns UTXO discovery/caching, miner owns claim orchestration, dialog is a thin presenter.
- The Rust FFI surface (`compute_nullifier`, `compute_address_hash_hex`, `compute_merkle_positions`, `decode_leaf_*`, `generate_proof`, `aggregate_proofs`) is small, focused, and matches the CLI's structure. Constants (`MAX_PROOFS_PER_BATCH = 16`, `VOLUME_FEE_BPS = 10`) are centralized in Rust.
- Cache file path uses a short hash prefix — good privacy practice for cache filenames.
- The 16-way concurrency in proof generation is the right shape; the parallelism overhead concerns are real but secondary.
- The progress UI's phase mapping (`phase + 1` from UTXO → step in dialog) is consistent and the stepper looks great.
- The keychain entitlement and Podfile changes are reasonable build-system hygiene.

## Suggested order of operations

1. Fix #1, #2, #3, #4 — all four are real correctness issues that can lose user money or freeze the cache.
2. Decide explicitly on the mnemonic re-add (#5) and call it out in the PR description if intentional, or pull it into its own PR.
3. Pull telemetry (#6) out into a separate PR with explicit user consent.
4. DRY the hex/format helpers (#7) and stop swallowing RPC errors (#8).
5. Cancellation cleanup (#9) and `eagerError` (#10).
6. Then add the few unit tests ethan-crypto called for.

Net assessment: the architecture is solid and most of the work is good — the blockers are isolated to a few specific functions, and they're all fixable without restructuring. I would not merge in current state, but with the four blocking items resolved and the mnemonic/telemetry questions answered, this is close.

ethan-crypto (commented):
### What looks fixed from the prior review

The **transfer cache off-by-one** issue appears fixed. `_getChainHeight()` now throws instead of using a fabricated fallback height, and `getTransfersTo` now passes `cache.cachedUpToBlock` directly while the GraphQL query uses `height_gt`, avoiding the prior skipped `N + 1` block. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/c254644/quantus_sdk/lib/src/services/wormhole_utxo_service.dart)][2])

The **net amount accounting** also looks fixed. `netTotal` now accumulates each proof’s post-fee output, and `_generateLeafProof` returns `outputAmount * _scaleDownFactor`, matching the comment that this is the planck amount dispatched on-chain. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/c254644/miner-app/lib/src/services/wormhole_claim_service.dart)][3])

The **destination address validation** and **cancelled terminal state** are improved. The dialog now validates by calling `getAccountId32`, and the progress header distinguishes “Claim Cancelled” from “Claim Complete.” ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/c254644/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart)][4])

The **finalized/nullifier cache issue** is improved. The UTXO service now keeps unfinalized spent nullifiers in memory for the current call but only persists nullifiers whose block height is at or below `safeCutoff`. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/c254644/quantus_sdk/lib/src/services/wormhole_utxo_service.dart)][2])

### Blocking / high-priority findings

**1. Chain selection is not safely wired into UTXO discovery or cache keys.**

This is the biggest remaining correctness issue I see. The claim dialog reads the selected `ChainConfig` and passes only `chainConfig.rpcUrl` into `claimRewards`, but `WormholeClaimService` constructs a `WormholeUtxoService` with no selected chain context and calls `getUnspentTransfers` without passing the selected RPC or Subsquid URL. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/c254644/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart)][4])

Inside `WormholeUtxoService`, the GraphQL and RPC endpoint services are created internally, while the miner settings code logs the selected chain endpoints but leaves the actual endpoint-setting calls commented out. The configured chains also clearly differ across `dev`, `dirac`, and `planck`; notably, Planck’s configured Subsquid URL is local `127.0.0.1:4000/graphql`. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/c254644/quantus_sdk/lib/src/services/wormhole_utxo_service.dart)][2])

The cache keys make this worse: transfer and nullifier cache filenames are derived only from the wormhole address hash prefix, not from chain ID, genesis hash, RPC URL, or Subsquid URL. A user switching between dev/Dirac/Planck with the same wormhole address can reuse transfer/nullifier cache from the wrong chain, skip queries because `cachedUpToBlock` is from another network, or try to prove leaf indices from a different chain. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/c254644/quantus_sdk/lib/src/services/wormhole_utxo_service.dart)][2])

**Suggested fix:** make `WormholeUtxoService` take explicit `rpcUrl`, `subsquidUrl`, and a stable chain identifier, ideally genesis hash. Include that chain identifier in transfer/nullifier cache filenames. Also clear or namespace existing caches when chain changes.

**2. The UI still says “claimed/complete” after only mempool acceptance.**

The PR now explicitly says it does not wait for inclusion, and `_submitExtrinsic` calls `author_submitExtrinsic`, returning the pool-accepted transaction hash rather than watching for `InBlock`, `Finalized`, or `ExtrinsicFailed`. That can be an intentional product trade-off, but the UI still sets the success message to “transfers claimed” and shows “Claim Complete.” ([[GitHub](https://github.com/Quantus-Network/quantus-apps/pull/469)][1])

That wording is too strong. A transaction accepted into the pool can still fail to land, be dropped, or be invalidated by a reorg/validation condition. The PR text argues that later nullifier checks will discover failures, but the dialog currently presents the operation as completed. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/pull/469)][1])

**Suggested fix:** either use `author_submitAndWatchExtrinsic` and report success only after `InBlock`/`Finalized` with event decoding, or change the UI/result semantics to “submitted/accepted by node,” display tx hashes, and refresh nullifiers/balance afterward before saying anything was actually claimed.

**3. Cancellation can return while the abandoned claim flow is still running.**

The service intentionally races the whole flow against a cancel future, with comments saying in-flight proofs are “orphaned” and finish in the background. But `Future.any` does not cancel the losing operation, and the `finally` block disposes the RPC client as soon as the cancel path wins. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/c254644/miner-app/lib/src/services/wormhole_claim_service.dart)][3])

This is especially risky around submission. The batch loop checks cancellation before aggregation and before `_submitExtrinsic`, but if cancellation happens while `_submitExtrinsic` is already in flight, the UI can show “Cancelled” even though a batch may still be accepted by the pool. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/blob/c254644/miner-app/lib/src/services/wormhole_claim_service.dart)][3])

**Suggested fix:** once step 6 starts, either disable cancellation or change the terminal copy to “cancel requested; already-submitted batches may still land.” Also avoid disposing the RPC client until the underlying flow has actually stopped, or structure the flow so cancellation does not abandon live RPC/submission work.

### Medium-priority findings

**4. GitHub still flags hidden/bidirectional Unicode in `miner_balance_card.dart`.**

The files page still shows GitHub’s warning that `miner_balance_card.dart` contains hidden or bidirectional Unicode text. This should be cleaned before merge, especially in a security-sensitive wallet/miner flow. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/pull/469/files)][5])

**5. TelemetryDeck is unrelated to the claim feature and should be split or gated.**

The PR body calls out TelemetryDeck as unrelated and review-worthy, and the current `main.dart` starts `Telemetrydecksdk` with a hardcoded `appID` and `salt`; `pubspec.yaml` also adds the dependency. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/pull/469)][1])

I would split this into a separate PR or gate it behind explicit product/privacy approval and user-facing disclosure. It is not part of the claim path and makes the review harder.

**6. Mnemonic persistence needs explicit security/release sign-off.**

The PR re-enables mnemonic persistence and notes that macOS builds require the proper Keychain entitlement and provisioning profile. The implementation stores mnemonics through `SettingsService` using `FlutterSecureStorage`, which is directionally right, but this is a meaningful security posture change and should be reviewed separately from reward claiming. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/pull/469)][1])

**7. Automated regression tests are still needed.**

The PR test plan is manual: full testnet claim, cancellation, zero-unspent, network interruption, and signed/notarized macOS keychain access. Those are useful, but the risky fixes here should have automated coverage: chain-scoped caches, cache off-by-one, RPC-height failure not mutating cache, safe nullifier persistence, invalid destination validation, cancellation during submission, and submit-vs-inclusion UI semantics. ([[GitHub](https://github.com/Quantus-Network/quantus-apps/pull/469)][1])

### Merge recommendation

I would **not merge yet**. The previous cache and validation bugs are largely addressed, but the chain-selection/cache namespace issue is a real correctness blocker, and the “Claim Complete” semantics still overstate what the code actually verifies. After fixing those two areas, I would be comfortable with the current-head proof strategy as a deliberate retry-on-reorg trade-off, provided the UI and tests make that behavior explicit.

[1]: https://github.com/Quantus-Network/quantus-apps/pull/469 "Miner: wormhole claim rewards by n13 · Pull Request #469 · Quantus-Network/quantus-apps · GitHub"
[2]: https://github.com/Quantus-Network/quantus-apps/blob/c254644/quantus_sdk/lib/src/services/wormhole_utxo_service.dart "quantus-apps/quantus_sdk/lib/src/services/wormhole_utxo_service.dart at c25464493701078674e79417c9d89b4fcc17fc58 · Quantus-Network/quantus-apps · GitHub"
[3]: https://github.com/Quantus-Network/quantus-apps/blob/c254644/miner-app/lib/src/services/wormhole_claim_service.dart "quantus-apps/miner-app/lib/src/services/wormhole_claim_service.dart at c25464493701078674e79417c9d89b4fcc17fc58 · Quantus-Network/quantus-apps · GitHub"
[4]: https://github.com/Quantus-Network/quantus-apps/blob/c254644/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart "quantus-apps/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart at c25464493701078674e79417c9d89b4fcc17fc58 · Quantus-Network/quantus-apps · GitHub"
[5]: https://github.com/Quantus-Network/quantus-apps/pull/469/files "Miner: wormhole claim rewards by n13 · Pull Request #469 · Quantus-Network/quantus-apps · GitHub"

n13:
Only (1) above is valid, it's super hacky to create our own endpoints, we should use the app constants

We do not support different chains, it's only one chain supported, the rest is debug code - everything else is as designed. We also need to ship this and it's a bit of an experimental product.

PR #470, opened 2026-04-30
Improve currency implementation

dewabisma:
## Summary
- Add real exchange rates
- Add currency conversion in send flow

## Screenshots
- Input amount screen
Simulator Screenshot - iPhone 8 - 2026-04-30 at 19 14 49

- Review send screen
Simulator Screenshot - iPhone 8 - 2026-04-30 at 19 14 53


n13:
Now I have a complete picture of the PR. Let me put together a thorough review.

---

# PR #470 Review – Improve currency implementation

[PR #470](https://github.com/Quantus-Network/quantus-apps/pull/470) wires real exchange rates into the wallet and lets users enter the send amount in either QUAN or their selected fiat. Overall this is a really nice improvement — the input flow with the swap button feels much more usable, and the cache‑then‑fetch strategy is the right shape. Sharing what I found below, grouped by severity.

## What I liked

- Clear separation of concerns: rates fetching/caching lives in `currency_display_provider.dart`, formatting still lives in `_toFiatDisplayState`, and `ExchangeRateService` is now a thin pure converter.
- `quanToUsdRate` is factored out as a constructor field so a future QUAN price feed can be wired in without touching call sites.
- The `_isUpdatingProgrammatically` guard around programmatic text changes in `input_amount_screen.dart` is the right pattern to prevent feedback loops with the controller listener.
- `txAmountDisplayProvider` gaining a `maxDecimals` parameter (used as `4` in send/review) keeps the amount you typed faithful in the review screen — nice touch.
- Reuse of `AmountDisplayWithConversion` in the review screen replaces hand‑rolled `Text`/style code — net +DRY.

## Critical

### 1. App will hard‑error on first launch when there's no cached rates yet

`mobile-app/lib/providers/currency_display_provider.dart`

```92:101:mobile-app/lib/providers/currency_display_provider.dart
final exchangeRateServiceProvider = Provider<ExchangeRateService>((ref) {
  final ratesAsync = ref.watch(exchangeRatesProvider);
  final settings = ref.read(settingsServiceProvider);

  return ratasync.when(
    data: (rates) => ExchangeRateService(rates: rates),
    loading: () => ExchangeRateService(rates: _readRatesCacheAnyAge(settings)),
    error: (_, _) => ExchangeRateService(rates: _readRatesCacheAnyAge(settings)),
  );
});
```

`_readRatesCacheAnyAge` throws (`Exception('No existing cached exchange rates!')`) when SharedPreferences has no entry — which is exactly the case on a fresh install. Trace:

- Cold start, no cache → `_readRatesCache` returns `null`, async fetch begins.
- For the brief window where `exchangeRatesProvider` is `loading`, `loading: ()` runs → `_readRatesCacheAnyAge` throws → `exchangeRateServiceProvider` itself errors → every consumer (`balanceDisplayProvider`, `txAmountDisplayProvider`, the input screen, etc.) sees an error.
- If the network call also fails (offline first launch), it stays broken.

Two fixes worth considering:

1. Restore the static `fallbackRates` map on `ExchangeRateService` (the docstring even refers to it: `Ultimate fallback: ExchangeRateService.fallbackRates`) and have `_readRatesCacheAnyAge` return it when nothing is cached.
2. Or have `exchangeRatesProvider` itself seed the cache with the fallback before returning, so the loading window never has an empty cache.

Either way, the provider must not throw on fresh install.

### 2. `data.cast<String, double>()` will TypeError when the API returns a whole number

```526:543:quantus_sdk/lib/src/services/taskmaster_service.dart
  Future<Map<String, double>> getExchangeRates() async {
    final http.Response response = await http.get(
      _exchangeRatesEndpoint,
      headers: {'Content-Type': 'application/json'},
    );
    if (response.statusCode != 200) {
      throw Exception('Exchange rates request failed with status: ${response.statusCode}. Body: ${response.body}');
    }

    final Map<String, dynamic>? responseBody = jsonDecode(response.body);
    final Map<String, dynamic>? data = responseBody?['data'];

    if (data == null) {
      throw Exception('Exchange rates not found!');
    }

    return data.cast<String, double>();
  }
```

`Map.cast<K2, V2>()` returns a lazy view; the type check happens on read. So the first time the consumer iterates and pulls a value that JSON parsed as `int` (e.g. `"USD": 1`), it throws a `TypeError` rather than the explicit "no rates" exception. Suggest:

```dart
return data.map((k, v) => MapEntry(k, (v as num).toDouble()));
```

This also makes the conversion in the consumer (`Decimal.parse(v.toString())`) deterministic.

## Important

### 3. Silent `catch (_)` swallows parse failures (violates the project "fail early" rule)

```19:32:mobile-app/lib/providers/currency_display_provider.dart
Map<String, Decimal>? _readRatesCache(SettingsService settings) {
  final raw = settings.getString(_kRatesCacheKey);
  if (raw == null) return null;
  try {
    final decoded = jsonDecode(raw) as Map<String, dynamic>;
    final expiryUnix = decoded['expiry'] as int;
    final nowUnix = DateTime.now().millisecondsSinceEpoch ~/ 1000;
    if (nowUnix >= expiryUnix) return null;
    final ratesJson = decoded['rates'] as Map<String, dynamic>;
    return ratesJson.map((k, v) => MapEntry(k, Decimal.parse(v as String)));
  } catch (_) {
    return null;
  }
}
```

Same shape in `_readRatesCacheAnyAge` and the `try { ... } catch (_) { return _readRatesCacheAnyAge(settings); }` inside `exchangeRatesProvider`. Per the repo guideline ("never have a silent failure, all failures must produce error logging"), please at least `debugPrint` the error before returning `null` / falling through. Otherwise a corrupt cache entry is invisible forever.

### 4. `getRate` throws but the docstring promises a fallback

```19:26:mobile-app/lib/services/exchange_rate_service.dart
  /// Returns the exchange rate for [fiat] (units per 1 USD).
  /// Falls back to [fallbackRates] if [fiat] is absent from the live map.
  Decimal getRate(FiatCurrency fiat) {
    final rate = _rates[fiat.code];
    if (rate == null) throw Exception('Exchange rate not found for ${fiat.code}!');

    return rate;
  }
```

`fallbackRates` doesn't exist anywhere, and there is no fallback. Practically: if the user has `MYR` selected but the API response doesn't include `MYR`, `_fiatStringToQuan` and `_toFiatNumeric` both crash on every keystroke. Either:

- Implement the fallback (static map), or
- Update the docstring and have callers handle `StateError` (and prefer `StateError` over bare `Exception`).

### 5. `_setMax()` non‑flipped path is asymmetric with the flipped path

```184:204:mobile-app/lib/v2/screens/send/input_amount_screen.dart
  void _setMax() {
    final balance = ref.read(effectiveMaxBalanceProvider).value ?? BigInt.zero;
    final max = SendScreenLogic.calculateMaxSendableAmount(balance: balance, networkFee: _networkFee);
    final isFlipped = ref.read(isCurrencyFlippedProvider);
    if (!isFlipped) {
      _amountController.text = _fmt.formatBalance(
        max,
        maxDecimals: AppConstants.decimals,
        addThousandsSeparators: false,
      );
    } else {
      _isUpdatingProgrammatically = true;
      try {
        _amountController.text = _quanToFiatString(max);
      } finally {
        _isUpdatingProgrammatically = false;
      }
      setState(() => _amount = max);
      if (max > BigInt.zero) _fetchFee();
    }
  }
```

In the QUAN branch you rely on `_onAmountChanged` (triggered by the controller setter) to update `_amount` and call `_fetchFee`; in the fiat branch you do it manually. That works today but is fragile — anyone moving lines around will reintroduce a double‑update or an unintended round‑trip through `_fiatStringToQuan(_quanToFiatString(max))` which can drift due to division. Suggest always wrapping with `_isUpdatingProgrammatically` and explicitly setting `_amount = max` + calling `_fetchFee()` in both branches.

### 6. Hidden balance + flipped mode appends `QUAN` to the masked text

```282:288:mobile-app/lib/providers/currency_display_provider.dart
        if (!isHidden) {
          data = data.copyWith(primaryAmount: withSignPrefix ? '$prefix${data.primaryAmount}' : data.primaryAmount);
        }

        if (!withQuanSymbol && isFlipped) {
          data = data.copyWith(secondaryAmount: '${data.secondaryAmount} ${AppConstants.tokenSymbol}');
        }
```

The QUAN‑symbol appending isn't gated by `isHidden`, so the secondary line renders as `- - - - - QUAN` when hiding a flipped balance. Add an `&& !isHidden` guard.

## Minor / nits

- `int maxDecimals` is misleading for the fiat path — `_toFiatNumeric` uses `toStringAsFixed(maxDecimals)`, which pads to *exactly* that many digits, so a USD balance shows `$100.000` instead of `$100.00` once `balanceDisplayProvider` passes `maxDecimals: 3`. Either rename to `fiatDecimals` and document the difference, or trim trailing zeros for fiat. Same caveat for `_quanToFiatString` which always emits 4 d.p. (`1.5` displays as `1.5000` after pressing Max in fiat mode).
- `const _kCacheTtlSeconds = 86400; // 24 hours` would read better as `const _kCacheTtl = Duration(hours: 24);` and then `_kCacheTtl.inSeconds` where needed.
- DRY: `_readRatesCache` and `_readRatesCacheAnyAge` repeat the JSON → `Map<String, Decimal>` decoding. Extract the inner parse into one helper that the two wrappers share.
- The `getExchangeRates` endpoint is `/exchange-rate` (singular) while the SDK method/field are plural — sanity‑check that matches the backend spec to avoid a 404 surprise.
- Stray docstring: `/// Falls back to [fallbackRates] for any code not present.` in `exchange_rate_service.dart` should be removed (or implemented).
- No tests added for the new caching strategy or the fiat ⇄ QUAN conversion math; would be very valuable since the conversion is now load‑bearing in the send flow.
- Commit 3 ("chore: formatting") sweeps 25 unrelated files — useful but ideally a separate PR. It also drops a trailing newline in `scaffold_base.dart` which is noise.
- `_fetchFee` is invoked on every keystroke inside `_onAmountChanged` (both before and after this PR) — not introduced here, but with fiat input it's now a network call per character of fiat input. Worth a debounce as a follow‑up.
- Style nit: in `amount_display_with_conversion.dart`, the `fontWeight: FontWeight.w400` override on top of `text.extraLargeTitle` is a magic value — prefer a named style on `AppTextTheme` so it can flex globally.

## Suggested follow‑ups

1. Land the fallback story before merge (issue 1 is a release blocker for offline first‑run).
2. Tighten `getExchangeRates` parsing (issue 2).
3. Decide whether `getRate` falls back or throws, and align doc + code (issue 4).
4. Add at least one unit test for `_writeRatesCache` / `_readRatesCache` round‑trip + an `ExchangeRateService.convert` test for a non‑USD currency.

Happy to dig deeper on any of these or pair on the fallback design if helpful.

n13:
need to fix most of these, even the maxDecimals nit...

n13:
Re-reviewed PR #470 against the latest commit (`7905d4e1 fix: PR review issues`). Great turnaround — every critical and important item from the previous round is meaningfully addressed. Quick rundown of what landed plus a few small things still worth a look.

## What got fixed

**1. Hard-error on first launch with no cache** — fixed correctly. `_readRatesCacheAnyAge` now returns `ExchangeRateService.fallbackRates` instead of throwing, and `ExchangeRateService.fallbackRates` is reintroduced as a real static map (USD, MYR, IDR, JPY, EUR, GBP). `exchangeRatesProvider`'s `catch` also routes to the same helper, so offline first-run is safe.

```50:56:mobile-app/lib/providers/currency_display_provider.dart
Map<String, Decimal> _readRatesCacheAnyAge(SettingsService settings) {
  final raw = settings.getString(_kRatesCacheKey);
  if (raw == null) return ExchangeRateService.fallbackRates;
```

**2. `data.cast<String, double>` TypeError** — fixed. Now eagerly converts via `(v as num).toDouble()`:

```542:542:quantus_sdk/lib/src/services/taskmaster_service.dart
    return data.map((k, v) => MapEntry(k, (v as num).toDouble()));
```

And the consumer side does `Decimal.parse(v.toString())` which is deterministic against the `double` map. Endpoint also changed to plural `/exchange-rates` — nice.

**3. `getRate` fallback** — actually implemented now, matches the docstring:

```34:38:mobile-app/lib/services/exchange_rate_service.dart
  Decimal getRate(FiatCurrency fiat) {
    final rate = _rates[fiat.code] ?? fallbackRates[fiat.code];
    if (rate == null) throw Exception('Exchange rate not found for ${fiat.code}!');
    return rate;
  }
```

**4. `_setMax()` symmetry** — both branches now go through the programmatic guard and set `_amount`/`_fetchFee` consistently:

```183:198:mobile-app/lib/v2/screens/send/input_amount_screen.dart
  void _setMax() {
    final balance = ref.read(effectiveMaxBalanceProvider).value ?? BigInt.zero;
    final max = SendScreenLogic.calculateMaxSendableAmount(balance: balance, networkFee: _networkFee);
    final isFlipped = ref.read(isCurrencyFlippedProvider);
    _isUpdatingProgrammatically = true;
    try {
      _amountController.text = isFlipped
          ? _quanToFiatString(max)
          : _fmt.formatBalance(max, maxDecimals: AppConstants.decimals, addThousandsSeparators: false);
    } finally {
      _isUpdatingProgrammatically = false;
    }
    setState(() => _amount = max);
    if (max > BigInt.zero) _fetchFee();
  }
```

**5. Hidden balance + flipped → "QUAN" leaking into masked text** — gated by `!isHidden`:

```279:281:mobile-app/lib/providers/currency_display_provider.dart
        if (!withQuanSymbol && isFlipped && !isHidden) {
          data = data.copyWith(secondaryAmount: '${data.secondaryAmount} ${AppConstants.tokenSymbol}');
        }
```

**6. Fiat decimals nit** — fully addressed. `FiatCurrency` now carries a `decimals` field (USD/EUR/GBP/MYR=2, IDR/JPY=0), `_toFiatNumeric` uses `fiat.decimals`, and `_quanToFiatString` does the same — so a USD balance is `$100.00` and IDR is `Rp17,337` rather than padded weirdness. The provider parameter is also renamed `quanDecimals`, which makes intent clear.

**7. DRY + Duration + tests** — `_parseRatesMap` helper extracted, `_kCacheTtlDuration = Duration(hours: 24)`, and `mobile-app/test/unit/exchange_rate_service_test.dart` covers `getRate` (with fallback), `convert`, `quanRawToFiat`, `fiatToQuanRaw`, plus a round-trip inverse test. 

**8. Fee fetch debounce** — `Debouncer(delay: 500ms)` wired into `_onAmountChanged` via the new shared `mobile-app/lib/shared/utils/debouncer.dart`. Cleanly reusable.

## Remaining things worth a quick pass

**a) Silent catches still present in two spots.** `_readRatesCache` was upgraded to `debugPrint` the parse failure (great), but `_readRatesCacheAnyAge` and the outer `exchangeRatesProvider` `catch` are still bare. Per the project "fail early / always log" rule, would be nice to mirror the same `debugPrint` so a corrupt cache or a network outage isn't invisible:

```86:90:mobile-app/lib/providers/currency_display_provider.dart
    return rates;
  } catch (_) {
    return _readRatesCacheAnyAge(settings);
  }
});
```

```54:57:mobile-app/lib/providers/currency_display_provider.dart
    return _parseRatesMap(decoded['rates'] as Map<String, dynamic>);
  } catch (_) {
    return ExchangeRateService.fallbackRates;
  }
```

**b) Misleading `getRate` test.** This test name claims to assert throwing, but the body asserts `returnsNormally` for USD (which is in `fallbackRates`), so the throw path is never actually exercised:

```31:35:mobile-app/test/unit/exchange_rate_service_test.dart
    test('throws when the currency is absent from both live and fallback rates', () {
      final serviceWithEmpty = ExchangeRateService(rates: {});
      // JPY is in fallbackRates, so we expect no throw for standard currencies.
      // A made-up code should throw.
      expect(() => serviceWithEmpty.getRate(FiatCurrency.values.firstWhere((c) => c.code == 'USD')), returnsNormally);
    });
```

Because `FiatCurrency` is a closed enum and every case exists in `fallbackRates`, the throw path is now unreachable through the type system. Either:
- delete the test (since the throw is a defensive guard against future code-only currencies), or
- make `getRate` accept a `String` (or test a `FiatCurrency` whose code you've stripped from a custom map) so the throw is actually covered.

The current test as-written is misleading — its name says one thing and it asserts the opposite.

**c) Round-trip test caveat.** `quanRawToFiat ↔ fiatToQuanRaw are inverses` only round-trips losslessly when the rate cleanly divides; with `1.5 QUAN @ 4 MYR` it works because `fiat = 6 MYR` is exact, but a real-world rate like 3.97 would *not* be lossless. Worth adding one explicit non-clean-rate test (e.g. `MYR=Decimal.parse('3.97')`) to lock in the expected rounding behavior — or at least drop a comment that round-trip equality only holds for clean rates.

**d) Tiny consistency nit** in `_fiatStringToQuan`:

```176:185:mobile-app/lib/v2/screens/send/input_amount_screen.dart
  BigInt _fiatStringToQuan(String fiatText) {
    if (fiatText.isEmpty) return BigInt.zero;
    final sanitized = fiatText.replaceAll(',', '.');
    final fiatDecimal = Decimal.tryParse(sanitized);
    if (fiatDecimal == null) return BigInt.zero;
```

The `replaceAll(',', '.')` only handles single-decimal-mark locales. If `DecimalInputFilter()` already normalises this, the line is dead; if not, IDR/JPY users typing `1.234,56`-style input will still misparse. Worth confirming the input formatter's contract — not blocking.

**e) Optional polish.** The `fontWeight: FontWeight.w400` magic value on top of `text.extraLargeTitle` in `amount_display_with_conversion.dart` is unchanged — still a minor candidate for a named style on `AppTextTheme`. Truly minor.

## Verdict

This is in great shape — the critical correctness and offline-startup risks are gone, the conversion math has tests, and the input flow is more robust with the debounce + symmetric `_setMax`. I'd LGTM after the two silent catches get a `debugPrint` and the misleading `getRate` test is either deleted or made to actually test the throw.

Nice work iterating on this!

-> not really tiny JPY won't parse? 

Also I think it will screw up USD you could have 1,000 (1000) -> 1.0 
Please fix fiat to string so it works... 

1.234,5 
1,000,000 (USD)
123,45 (IDR)

n13:
So when a USD user (English locale, comma = thousands separator) types 1,000 meaning one thousand dollars:

DecimalInputFilter happily accepts 1,000 (matches the regex — one comma followed by ≤12 digits).
_fiatStringToQuan runs replaceAll(',', '.') → "1.000".
Decimal.tryParse("1.000") → 1.0.
fiatToQuanRaw converts 1 USD → ~1 QUAN raw.
The user thinks they're sending $1,000 worth of QUAN. They're actually sending $1 worth. There is no error, no warning, no rejection. The field even continues to display 1,000 while _amount holds 1/1000th of what they intended.

The _amount then propagates to the review screen and the actual transaction.

n13:
people can also paste into this field I think.. need to somehow clarify thousands separators, in India they even have 100s separators (looks super weird)

n13:
I think it's pretty safe to use the user's current locale to do parsing and find out about separators... if they paste another locale format then it's ok to fail, since it doesn't make sense...

we should have unit tests for a bunch of different locales...

n13:
Can we make the conversions throw errors on invalid input? 

Then the UX can handle invalid inputs...
instead of returning 0

## PR #470 Review — round 3 (`2e858540`)

This round addresses every blocking item from the previous review and adds genuinely impressive locale plumbing. The 102‑test `locale_number_handling_test.dart` is the highlight — it locks down the exact "1,000 silently becomes 1.0" footgun n13 called out, plus a swathe of cross‑locale edge cases (US, ID, DE, FR, GB, JP, MY). All tests pass locally.

### What got fixed since last round

- **Locale-aware parsing** — `_fiatStringToQuan` now goes through `LocaleNumberConfig.normalize()` instead of the broken `replaceAll(',', '.')`. So a US user typing `1,000` parses as 1000 USD (not 1 USD), and an Indonesian user typing `1.000` parses as 1000 IDR. Both directions tested.
- **`DecimalInputFilter` is locale-aware** — distinguishes typing (1 char) from paste (>1 char). Typing accepts either `,` or `.` and converts to the locale's decimal (handles iOS keyboards that show the "wrong" character). Paste applies the full locale rules (strip grouping). Nicely thought through, and `pos_amount_screen` reuses it.
- **Silent catches now log** — the outer `exchangeRatesProvider` catch and `_readRatesCacheAnyAge` both `debugPrint(...)` now. ✓
- **Misleading `getRate` test** — replaced with a clean `'falls back to fallbackRates for a missing live rate'` test that actually exercises the fallback path.
- **Cache TTL is server-driven** — replaced the local 24h `Duration` with the API's `time_next_update_unix`. Smarter than guessing.
- **`ExchangeRatesResult` model** — `(v as num).toDouble()` is eager, so an integer rate from the API no longer triggers a lazy `TypeError` downstream.
- **Chain-history merge cleanup** is correct — the deleted `fetchScheduledReversibleTransfers` public wrapper had no remaining external callers (verified), so no functionality was lost.

### Remaining issues, by severity

**Important (worth fixing before merge)**

1) **Stray `print` of cache payload to release logs** — `currency_display_provider.dart` line 68:

```68:68:mobile-app/lib/providers/currency_display_provider.dart
  print('writing exchange rates cache: $payload');
```

The same file uses `debugPrint` everywhere else in this PR — this `print` is inconsistent and will leak the entire rates payload + expiry to production logs (it runs every successful fetch). Make it `debugPrint(...)` or remove.

2) **`_fiatStringToQuan` silently returns `BigInt.zero` on parse failure** — violates "fail early / always log":

```196:205:mobile-app/lib/v2/screens/send/input_amount_screen.dart
  BigInt _fiatStringToQuan(String fiatText) {
    if (fiatText.isEmpty) return BigInt.zero;
    final normalized = _localeConfig.normalize(fiatText);
    final fiatDecimal = Decimal.tryParse(normalized);
    if (fiatDecimal == null) return BigInt.zero;
```

After the locale `normalize()` already succeeded, the only way `Decimal.tryParse` returns null is genuinely malformed input (e.g. partial paste, `1.`, `,`). Today this is invisible — the user sees nothing and `_amount` becomes 0. A `debugPrint('Failed parsing fiat input "$fiatText" → "$normalized"');` (and matching to either the `parseAmount` or `formatBalance` style elsewhere) is the minimum.

**Minor**

3) **Cache expiry not defended against bad `time_next_update_unix`** — if the upstream returns 0 or a past timestamp (server bug), `_readRatesCache` reports the cache as expired on every cold start and we re‑hit the network forever. Worth a floor like `expiryUnix > nowUnix + minTtl` before persisting, or skipping the persist altogether.

4) **Stale doc** — `NumberFormattingService.parseAmount` still says "When [localeConfig] is provided" (`number_formatting_service.dart` lines 63‑65), but `localeConfig` is now a required constructor field. Same file's class docstring on `formatBalance` example uses `1234500000000 -> "1.2345"` which is fine; just the `parseAmount` doc is outdated.

5) **`fontWeight: FontWeight.w400` magic value** still present (`amount_display_with_conversion.dart` line 68). This was a minor nit in round 2 — still worth pulling into a named style on `AppTextTheme` so it's themable, but truly cosmetic.

6) **`localeNumberConfigProvider` snapshots `Platform.localeName` once** — Riverpod's `Provider` won't rebuild if the user changes their device locale mid‑session (would require an app restart). Acceptable as the common case, but worth a `// caveat:` comment in the provider so a future maintainer doesn't trip on it.

7) **Defensive nit in `LocaleNumberConfig.normalize`** — no guard against `groupingSeparator == decimalSeparator`. With `intl`'s data this never happens for real locales, but a cheap `&& groupingSeparator != decimalSeparator` on the strip step would make the helper safe under any input.

**Optional follow‑ups**

8) The round‑trip test in `exchange_rate_service_test.dart` still uses clean‑divisor rates (`MYR=3.97`, `1.5 QUAN` → exact). The earlier feedback about a non‑clean rate (e.g. `Decimal.parse('3.971')`) test, or at minimum a `// round-trip equality holds only for clean rates` comment, would lock in expected rounding behavior. Not a blocker — the math itself is exercised by `quanRawToFiat` / `fiatToQuanRaw` individually.

9) Nice to have: a single widget‑level test on `InputAmountScreen` that types into the fiat field and asserts `_amount` ends up correct end‑to‑end. The unit pieces are now well‑tested in isolation; one integration test would prove the whole pipeline.

### Verdict

Strong round. Items 1 and 2 are the only ones I'd push for before merging — they're 5‑line changes and both are direct violations of the project's "fail early, always log" rule. Everything else is polish or future‑maintainer ergonomics. Once `print` becomes `debugPrint` and the `_fiatStringToQuan` silent‑zero gets a log, this is good to land.

Earlier reviews of this PR: [Initial review of currency PR](e3a32c1f-fc18-4c9a-b2cd-83b1a0f5b203), [Re-review of currency PR](7a6b4f2b-0c4a-4b1e-9e6f-d4f3a8b8c712).

n13:
I have everything I need. Let me write up the round 4 review.

---

## PR [#470](https://github.com/Quantus-Network/quantus-apps/pull/470) Review — round 4 (`1014c8f2`)

Both blockers from round 3 are addressed in this commit. The throw-and-handle pattern lines up nicely with what was requested ("make the conversions throw errors on invalid input… UX can handle invalid inputs"). I empirically verified the throw is mostly defensive (see below), so the toaster won't fire during normal typing — good.

### What got fixed since round 3

**1. Stray `print` of cache payload** — gone.

```62:71:mobile-app/lib/providers/currency_display_provider.dart
Future<void> _writeRatesCache(
  SettingsService settings,
  Map<String, Decimal> rates,
  int timeNextUpdateUnix,
) async {
  final payload = {
    'expiry': timeNextUpdateUnix,
    'rates': rates.map((k, v) => MapEntry(k, v.toString())),
  };
  
  await settings.setString(_kRatesCacheKey, jsonEncode(payload));
}
```

(Tiny: the diff left a stray blank line where the print was — easy to drop in a future formatter pass.)

**2. `_fiatStringToQuan` no longer silently returns zero on parse failure** — now throws and the caller handles it.

```200:214:mobile-app/lib/v2/screens/send/input_amount_screen.dart
  BigInt _fiatStringToQuan(String fiatText) {
    if (fiatText.isEmpty) return BigInt.zero;
    final normalized = _localeConfig.normalize(fiatText);
    final fiatDecimal = Decimal.tryParse(normalized);
    if (fiatDecimal == null) {
      throw Exception('Invalid fiat input: $fiatText');
    }

    final xRate = ref.read(exchangeRateServiceProvider);
    final selectedFiat = ref.read(selectedFiatCurrencyProvider);
    return xRate.fiatToQuanRaw(fiatDecimal, selectedFiat, AppConstants.decimals);
  }
```

```122:139:mobile-app/lib/v2/screens/send/input_amount_screen.dart
  void _onAmountChanged() {
    if (_isUpdatingProgrammatically) return;
    final isFlipped = ref.read(isCurrencyFlippedProvider);
    if (isFlipped) {
      try {
        final convertedAmount = _fiatStringToQuan(_amountController.text);
        setState(() => _amount = convertedAmount);
      } catch (e) {
        context.showErrorToaster(message: e.toString());
        return;
      }
    } else {
      final formattingService = ref.read(numberFormattingServiceProvider);
      final parsed = formattingService.parseAmount(_amountController.text);
      setState(() => _amount = parsed ?? BigInt.zero);
    }
    if (_amount > BigInt.zero) _feeDebouncer.run(_fetchFee);
  }
```

I sanity‑checked toaster spam by probing `Decimal.tryParse` on the typing intermediate states the input filter actually allows through:

| input | `Decimal.tryParse` |
| --- | --- |
| `"1"` | `1` |
| `"1."` | `1` |
| `"1.5"` | `1.5` |
| `"0."` | `0` |
| `"0"` | `0` |
| `"."` | `null` |
| `","` | `null` |

`DecimalInputFilter`'s regex (`^(0\|([1-9]\d*))(\.\d{0,12})?$`) only ever lets through strings that `Decimal.tryParse` accepts, so the `throw` is effectively defensive — it should never fire on normal interactive typing. The toaster will only surface for genuinely malformed text that somehow bypasses the filter (or a future filter regression). 

### Remaining items worth a quick pass

**Important — small, but two project‑rule violations**

**a) Catch is too broad and not logged** in `_onAmountChanged`. Per the repo "fail early / always log" rule, the bare `catch (e)` should `debugPrint` before showing the toaster, so a future regression that fires this path is observable:

```dart
} catch (e, stack) {
  debugPrint('Fiat→QUAN parse failed: $e\n$stack');
  context.showErrorToaster(message: 'Please enter a valid amount');
  return;
}
```

**b) Toast message leaks Java‑speak to the user.** `e.toString()` produces `"Exception: Invalid fiat input: 1abc"` — that's developer text, not a user‑facing message. Either swap to a friendly string (above), or use a typed exception class and translate it at the boundary:

```dart
class InvalidFiatInputException implements Exception { ... }
```

This also lets the `catch` be narrowed to `on InvalidFiatInputException` so unrelated failures (provider read, etc.) don't get swallowed as "Invalid fiat input" toasts.

**Minor (carried over from round 3 — none are blockers)**

These were all flagged previously and are still open. None are merge‑blocking, so up to you whether to land them in this PR or as a small follow‑up:

- **Cache expiry not defended against bad `time_next_update_unix`** — if the upstream returns `0` or a past timestamp, `_readRatesCache` reports the cache as expired on every cold start and we re‑hit the network forever. Worth `if (timeNextUpdateUnix > nowUnix + minTtl)` before persisting.
- **Stale doc** in `NumberFormattingService.parseAmount` (lines 63–65) still says *"When [localeConfig] is provided"* — but `localeConfig` is now a required constructor field.
- **`FontWeight.w400` magic value** still in `amount_display_with_conversion.dart:68` — would be cleaner as a named style on `AppTextTheme`.
- **`localeNumberConfigProvider` snapshots `Platform.localeName` once** — won't pick up a mid‑session locale change; worth a `// caveat: locale only re-read on app restart` comment so the next maintainer doesn't trip on it.
- **`LocaleNumberConfig.normalize`** has no guard for `groupingSeparator == decimalSeparator`. With intl's data this never happens for real locales, but a cheap `&& groupingSeparator != decimalSeparator` on the strip step makes the helper safe under any input.
- **Round‑trip test in `exchange_rate_service_test.dart`** still uses clean‑divisor rates only. A single `Decimal.parse('3.971')` test (or a `// round-trip equality holds only for clean rates` comment) would lock in expected rounding behavior.
- **No widget‑level integration test** on `InputAmountScreen` — would prove the fiat input → `_amount` pipeline end‑to‑end now that the unit pieces are well‑tested.

### Verdict

Both round‑3 blockers are gone and the throw approach is correctly wired to the UX. The two "important" items above are both ~5‑line changes and both are direct violations of the project's "fail early, always log" rule — I'd push for those before merging. Everything else is polish.

Nice iteration.

---

Earlier reviews of this PR: [Initial review of currency PR](9ea594fe-d501-4955-8f41-70cc19f53e52), [Re-review of currency PR](a4309fed-da77-4374-965d-263623e85a65), [Round 3 review of currency PR](b8a312ec-0d88-4550-b2ca-630a7a538004).

n13:
Please check see which ones of the above are worth fixing

n13:
Actually just fixing things, this is getting too long

n13:
All 111 tests pass. Here's my round-5 review.

## PR #470 Review — round 5 (b9949f5)

The two round-4 blockers are both addressed cleanly, and the typed‑exception refactor is the right call. The diff is small, focused, and well‑tested. Headline:

| round-4 item | status |
|---|---|
| Stray `print(...)` in cache write | gone |
| Bare `catch (e)` + `e.toString()` toast | typed `InvalidNumberInputException`, `debugPrint(... + stack)`, friendly "Please enter a valid amount" |
| Cache TTL not defended against bad `time_next_update_unix` | `_kMinCacheTtlSeconds = 60` guard with skip+log |
| Stale `parseAmount` doc | rewritten |
| `FontWeight.w400` magic value | named `conversionAmountPrimary` style on `AppTextTheme` |
| `localeNumberConfigProvider` snapshot caveat | comment added |
| `normalize` defensive guard for `groupingSeparator == decimalSeparator` | added + test |
| Round-trip lossy test | added with `Decimal.parse('3.971')`, asserts ≤ 1 raw-unit drift |

What's especially nice in this round:

- `LocaleNumberConfig.parseDecimal(...)` is a clean abstraction — both `NumberFormattingService.parseAmount` and `_fiatStringToQuan` go through it now, so the trailing‑decimal tolerance and locale rules live in exactly one place. Good DRY.
- The exception payload (`rawInput` + `normalized`) is exactly the shape a debug log wants; `'1.2.3,nope' → normalized "123.nope"` test locks that in.
- Trailing-decimal tolerance for mid-typing (`"1." → 1`, `"1," → 1`) is the right call — it avoids false toaster fires while the user is still typing.

### Suggestions (none are merge blockers)

**1. `_onAmountChanged` leaves `_amount` stale on parse failure** — `mobile-app/lib/v2/screens/send/input_amount_screen.dart` lines 122-140

```125:140:mobile-app/lib/v2/screens/send/input_amount_screen.dart
    if (isFlipped) {
      try {
        final convertedAmount = _fiatStringToQuan(_amountController.text);
        setState(() => _amount = convertedAmount);
      } on InvalidNumberInputException catch (e, stack) {
        debugPrint('Fiat→QUAN parse failed: $e\n$stack');
        context.showErrorToaster(message: 'Please enter a valid amount');
        return;
      }
    } else {
      final formattingService = ref.read(numberFormattingServiceProvider);
      final parsed = formattingService.parseAmount(_amountController.text);
      setState(() => _amount = parsed ?? BigInt.zero);
    }
```

The QUAN branch resets `_amount = BigInt.zero` on a null parse; the fiat branch leaves the previous `_amount` in place after the toaster. With `DecimalInputFilter` upstream this is unreachable in practice (the throw is purely defensive), but if it ever does fire, the user can still hit Send with whatever stale `_amount` was previously valid — text and `_amount` would be out of sync. One‑liner fix:

```dart
} on InvalidNumberInputException catch (e, stack) {
  debugPrint('Fiat→QUAN parse failed: $e\n$stack');
  setState(() => _amount = BigInt.zero);
  context.showErrorToaster(message: 'Please enter a valid amount');
  return;
}
```

**2. `_fiatStringToQuan`'s catch is narrower than the call surface** — same file, lines 200-210

```204:210:mobile-app/lib/v2/screens/send/input_amount_screen.dart
  BigInt _fiatStringToQuan(String fiatText) {
    if (fiatText.isEmpty) return BigInt.zero;
    final fiatDecimal = _localeConfig.parseDecimal(fiatText);
    final xRate = ref.read(exchangeRateServiceProvider);
    final selectedFiat = ref.read(selectedFiatCurrencyProvider);
    return xRate.fiatToQuanRaw(fiatDecimal, selectedFiat, AppConstants.decimals);
  }
```

`fiatToQuanRaw` calls `getRate(fiat)`, which throws a generic `Exception('Exchange rate not found ...')` when neither the live map nor `fallbackRates` has the code. Today every `FiatCurrency` enum value exists in `fallbackRates` so this can't fire — but the `on InvalidNumberInputException` catch will _miss_ that path if anyone adds a 7th currency without a fallback entry, and it'll crash the listener silently. Either:
- broaden the catch (`catch (e, stack)` with `debugPrint` + the same friendly toast), or
- give `getRate` a typed exception (e.g. `MissingExchangeRateException`) and catch both.

**3. Minor — cache the locale config field instead of re‑reading per access** — `_localeConfig` is a getter that calls `ref.read(...)` every time. Easy to make it a `late final` field initialized in `initState` (matches how `_feeDebouncer` and friends are stored). Tiny perf+clarity win.

### Carry-overs that are still optional (already noted in earlier rounds)

- A widget-level integration test on `InputAmountScreen` that types into the fiat field and asserts `_amount` end‑to‑end. The unit pieces are bulletproof now; an integration test would prove the wiring.
- Cross-locale paste ambiguity is a UX call, not a bug — a US user pasting `1.000` from an Indonesian source still silently means "1 with three decimals". Not solvable without a UX prompt; fine to defer.

### Verdict

Both round-4 blockers are gone, the typed-exception path is wired correctly, and the locale test coverage is comprehensive (111 tests, all green locally). The three items above are 1–3 line changes and not blocking; even just #1 (the stale `_amount` reset) would tighten the failure-mode contract enough that I'd LGTM after.

Nice iteration.

Earlier reviews of this PR: [Initial review of currency PR](e3a32c1f-fc18-4c9a-b2cd-83b1a0f5b203), [Re-review of currency PR](7a6b4f2b-0c4a-4b1e-9e6f-d4f3a8b8c712), [Round 3 review of currency PR](b8a312ec-0d88-4550-b2ca-630a7a538004), [Round 4 review of currency PR](9ea594fe-d501-4955-8f41-70cc19f53e52).

n13 (approved):

PR #482, opened 2026-05-09
refresh miner balance periodically and on mining new block

n13:
automatic balance refresh

- Balance refresh is expensive - we query the chain for all nullifiers etc
- Balance refreshes when we submit a block
- Slight delay to allow network to update
- Balance refreshes every 5 minutes in addition to that
- Add logging for sanity

image

### DIFF
--- a/miner-app/lib/main.dart
+++ b/miner-app/lib/main.dart
@@ -9,6 +9,7 @@ import 'features/setup/rewards_address_setup_screen.dart';
 import 'features/miner/miner_dashboard_screen.dart';
 import 'features/withdrawal/withdrawal_screen.dart';
 import 'src/services/binary_manager.dart';
+import 'src/services/log_file_sink.dart';
 import 'src/services/miner_wallet_service.dart';
 import 'src/services/mining_orchestrator.dart';
 import 'src/services/process_cleanup_service.dart';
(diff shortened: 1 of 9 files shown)


n13:
Tested, this works... mined a block, balance updated

before/after

image


image

miner-app/lib/main.dart
```diff
@@ -157,6 +158,13 @@ Future<void> main() async {
 
   // Note: Startup cleanup removed for simplicity
 
+  try {
+    final logsDir = await LogFileSink.getLogsDirectory();
```
dewabisma, line 162: What is this for?

dewabisma (commented):
Overall looks good to me, just wondering why we have that function call in main.dart

miner-app/lib/main.dart
n13: Logging. We didn't do any logging before! So now we have rotating logs, this is intializing the directory

dewabisma: Okay, so this is more about initializing logs rather than getting it to be used somewhere. IMO kind of misleading. It does not just getting log then.

n13: I mean we do print the path here, which is useful... ??!

n13: If variable is not used it's always a linter error anyway but this one is used, and I like the printout, because it tends to be very confusing to figure out where the logs are going...

dewabisma: Yeah the log is useful, I like it too. Just the method called get but it also do more than just get. Anyway, it's fine it we don't mind having side effect in the function

dewabisma (approved):
LGTM

miner-app/lib/main.dart
n13: I think a get with default is normal...

PR #490, opened 2026-05-19
Dependency cleanup

n13:
- Dependency cleanup
- Update some deps to new major versions
- Minor bumps for firebase
- Unify analysis options (linter) across the 3 projects
- Fix melos run format error - we now scope formatting to source files only, lib and test folders! 

Tested: App runs!


miner-app/lib/src/services/wormhole_claim_service.dart
```diff
@@ -382,8 +381,8 @@ class WormholeClaimService {
 
   static List<int> _encodeDigest(Map<String, dynamic> digest) {
     final logs = digest['logs'] as List<dynamic>? ?? [];
-    final output = scale.ByteOutput(256);
-    scale.CompactCodec.codec.encodeTo(logs.length, output);
+    final output = ByteOutput(256);
```
dewabisma, line 384: Where does this compact codec and byte output come from now?

mobile-app/lib/providers/notification_config_provider.dart
```diff
@@ -1,7 +1,7 @@
 import 'dart:async';
 import 'dart:convert';
 
-import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_riverpod/legacy.dart';
```
dewabisma, line 4: Breaking changes to riverpod v3 I guess

mobile-app/lib/services/deep_link_service.dart
```diff
@@ -5,7 +5,9 @@ import 'package:resonance_network_wallet/providers/account_associations_provider
 import 'package:resonance_network_wallet/providers/route_intent_providers.dart';
 
 final deepLinkServiceProvider = Provider<DeepLinkService>((ref) {
-  return DeepLinkService(ref);
+  final service = DeepLinkService(ref);
+  ref.onDispose(service.dispose);
```
dewabisma, line 9: why do we need this?

mobile-app/lib/app.dart
```diff
@@ -33,12 +33,6 @@ class _ResonanceWalletAppState extends ConsumerState<ResonanceWalletApp> {
     if (Platform.isAndroid) _referralService.checkPlayStoreReferralCode();
   }
 
-  @override
```
dewabisma, line 36: So recommended to use the ref on dispose instead of this approach?

quantus_sdk/lib/quantus_sdk.dart
```diff
@@ -73,6 +73,11 @@ export 'src/quantus_payload_parser.dart';
 export 'src/models/entrusted_account.dart';
 export 'src/models/display_account.dart';
 
+// Blockchain-infrastructure third-party re-exports. Apps consume these via
+// the SDK rather than depending on polkadart / bip39_mnemonic directly.
+export 'package:bip39_mnemonic/bip39_mnemonic.dart' show Mnemonic, Language;
+export 'package:polkadart/scale_codec.dart' show ByteOutput, CompactCodec;
```
dewabisma, line 79: Ohh I see it comes from this now for compact and btye output

dewabisma (approved):
LGTM!

miner-app/lib/src/services/wormhole_claim_service.dart
n13: this is just an import change - same code

mobile-app/lib/app.dart
n13: yeah

n13: the previous didn't work since ref gets disposed before dispose on app is called...

PR #492, opened 2026-05-20
Rust bridge update

n13:
- update flutter rust bridge
- update SPM migration - this failed the build, might be unrelated but good to have
- added fix_ios_build.sh in case it happens again

SPM error in Flutter see here: https://github.com/flutter/flutter/issues/162196


dewabisma (approved):
LGTM!

Not sure about that hacky fix, I didn't get any error but if it fix yours then GTG.

Also, please fix the CI error.

Seems the problem is the same as what I get, I solved it by running `melos bootstrap` locally then check the package.lock update in.

PR #503, opened 2026-06-01
fix fetching fee UX

dewabisma:
## Summary
Ensure button is disabled when fee is stale. Also show loader in the fetching fee section.


dewabisma (commented):
I think it's good now

PR #504, opened 2026-06-02
Cold wallet app

n13:
Cold wallet - installed on an unused phone in offline mode

This basically a hardware wallet that can go on any phone, and is used  with the Quantus Mobile Wallet as a signer.

Use any phone in offline mode, for security that phone shouldn't be used for anything else and be in airplane mode at all times.

- Import / Create wallet
- Manual entry of seed phrase since the cold wallet is not connected to any networks
- Show the QR code for the address (onboard to mobile app)
- Scan transaction QR code
- Show signature QR code
- Works only in offline mode

Simulator Screenshot - iPhone 17 - 2026-06-02 at 23 04 25
Simulator Screenshot - iPhone 17 - 2026-06-02 at 23 04 31
Simulator Screenshot - iPhone 17 - 2026-06-02 at 23 04 48

Simulator Screenshot - iPhone 17 - 2026-06-02 at 23 04 16

Simulator Screenshot - iPhone 17 - 2026-06-02 at 23 05 43
Simulator Screenshot - iPhone 17 - 2026-06-02 at 23 06 00


dewabisma (commented):
Maybe we can have a shared folder for both mobile app? Instead of duplicating the component and styling and everything since both need to be in same design system anyway. 

So, if we need to adjust design we don't have to do it twice.

Like the logo, Q with cold vibe :D

n13:
Ok we discussed and came to the conclusion and explanation: This is a cute app we want to have no development overhead for

I'd rather have it go off brand than having a constant dev mental overhead whenever we release the mobile app we also need to think about, check, and potentially fix this app - this would then mean it's no longer a cute app, but a real product. 

For now this signer app is internal to Quantus and will likely see no more than 10 users ever ;)

dewabisma (approved):
Okay look good, nothing serious found.

PR #507, opened 2026-06-05
Create proposal flow

dewabisma:
## Summary
Added proposal creation feature. This encompass the whole step of picking recipient, inputing amount, reviewing, and pending proposal submission.

We also show the proposal created event in propose account activity as to not confuse proposer by balance being deducted.

## Screenshots
- Multisig home
Simulator Screenshot - iPhone 8 - 2026-06-05 at 17 48 56

- Proposal details sheet
Simulator Screenshot - iPhone 8 - 2026-06-05 at 17 49 01

- Proposal details sheet
Simulator Screenshot - iPhone 8 - 2026-06-05 at 17 49 05

- Select recipient
Simulator Screenshot - iPhone 8 - 2026-06-05 at 17 49 15

- Input amount
Simulator Screenshot - iPhone 8 - 2026-06-05 at 17 49 20

- Review proposal
Simulator Screenshot - iPhone 8 - 2026-06-05 at 17 49 27

- After review done
Simulator Screenshot - iPhone 8 - 2026-06-05 at 17 49 30

---

> [!NOTE]
> **Medium Risk**
> Changes effective balance and background chain submission/reconciliation for proposals; approve/sign paths were removed, reducing co-signer risk but increasing reliance on polling and optimistic UI correctness.
> 
> **Overview**
> Adds an end-to-end **multisig transfer proposal** flow: recipient → amount (fee breakdown, fiat flip, member vs multisig balance checks) → review → **async on-chain submit** with optimistic **pending** rows, **indexer polling**, reconciliation into activity, and timeout/submit-failure toasts.
> 
> **Proposer-facing activity and balances** now treat proposal creation like other pending sends: pending proposals merge into transaction lists (with dedup when indexed), activity/detail sheets show “Proposing” / “Proposal created”, and **effective balance** reserves the proposer’s **member cost** (deposit + fees) while confirmation is in flight.
> 
> The **multisig home** switches to Receive + Propose actions and a unified **`MultisigActivitySection`** (open proposals pinned, recent merged feed). Proposal detail moves to a **read-only bottom sheet**; the previous **approve/cancel full-screen flow is removed**, with sign left disabled (“coming soon”). Copy and l10n shift toward “transfer” wording and explicit deposit vs proposal fee labels.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00deb424ac0abc1a8a32a997273f302871668882. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (changes requested):
## Updated review — HEAD `cdf7f8f2`

This is a **fresh re-review of the current head**, performed in an isolated detached worktree at `cdf7f8f2`, diffed against the **real PR base** `beast/discover-multisig-flow` (not `main`):

```
git diff origin/beast/discover-multisig-flow...HEAD
56 files changed, 3749 insertions(+), 1411 deletions(-)
```

14 commits since base, of which **5 are new** since the prior review at `00deb42`:
`d5288ce fix: redudancy` · `edf1226 fix: graphql query for multisig` · `2b38538 fix: circle dependency` · `889a022 feat: explicitly show failed fee fetch` · `cdf7f8f fix: reduce risk of double proposal creation`.

This is a **static read** (worktree has no `.dart_tool`/generated FRB bindings; nothing was built or run).

---

## Status of previously-raised issues

### Prior review (n13)

| # | Issue | Status |
|---|-------|--------|
| 1 | Nested submission retries → duplicate-proposal risk | ✅ **Resolved** |
| 2 | Fee-fetch fallback masks failure (fail-early) | ✅ **Resolved** |
| 3 | Circular import `multisig_proposal` ↔ `multisig_service` | ✅ **Resolved** |
| nit | `_onAmountFocusChanged` `Future.delayed(300ms)` + `ensureVisible` | ⬜ Not addressed (trivial) |
| nit | redundant `_palletConstants` / `_feeConstants` | ✅ Resolved |
| nit | test re-hardcodes `10000` step factor | ✅ Resolved |

**1 — Duplicate-proposal retries → Resolved.** `cdf7f8f2` removed the outer `maxRetries`/`attempt` recursion in `_submitProposalBackground`. It now issues a single `service.propose(...)` (which still retries internally in `submitExtrinsic`) and, on failure, removes the pending overlay and shows a `submitFailed` toast — no more 3×3 nonce-refetching attempts. See `mobile-app/lib/services/transaction_submission_service.dart:178-202`.

**2 — Fee fallback masks failure → Resolved.** `889a0227` deleted the catch-block that rebuilt a `networkFee: BigInt.zero` breakdown and set `_hasFee = true`. The catch now sets `_feeBreakdown = null; _hasFee = false; _feeFetchFailed = true;`, logs the error with stack, keeps submit disabled (`btnDisabled` includes `_feeFetchFailed`, line 272), and renders an explicit error + retry button. See `propose_amount_screen.dart:196-205, 522-546`. (One residual masking path remains — see N1.)

**3 — Circular import → Resolved.** `2b385381` moved `proposalCreationFeeFor` onto `MultisigProposal` (using the generated pallet `Constants()`) and dropped the `multisig_service.dart` imports from `multisig_proposal.dart`, `multisig_proposal_created_event.dart`, and `multisig_proposal_event.dart`. The dependency is now one-directional (service → model). See `multisig_proposal.dart:4,137-150` and `multisig_service.dart:296`. (Placed on the model rather than a constants/util as suggested — acceptable.)

### Cursor Bugbot

| # | Issue | Status |
|---|-------|--------|
| A | Done screen always shows `1/N` signatures | ⚠️ **Not resolved** (low–medium) |
| B | Success screen shown before submit finishes | 🟡 **Partially resolved** |
| C | Stale fee after skipped refetch | ✅ **Resolved** |
| D | Indexer timeout drops pending state without reconcile | ❌ **Not resolved** (medium) |

**A — Hardcoded `1/N` → Not resolved.** `propose_done_screen.dart:122` still renders `l10n.multisigSignaturesCount(1, msig.threshold)`. None of the 5 commits touched this file. The value is hardcoded and shown optimistically *before* on-chain confirmation. The repo's own fixtures/model treat the proposer as the first approval (`approvals: ['5Proposer']` → `approvalCount == 1` in `multisig_service_test.dart:340-352`, and `multisig_proposal_detail_sheet.dart:137` derives `approvalCount` for the real view), so `1/N` is *plausibly* consistent with the eventual indexed state — but it should be derived from confirmed state (or the copy softened), not assumed on a pre-confirmation screen.

**B — Success before submit → Partial.** `proposeTransfer` still returns immediately after enqueuing `unawaited(_submitProposalBackground(...))` (`transaction_submission_service.dart:159-168`), and the review screen `await`s it then navigates straight to the "submitted" done screen (`propose_review_screen.dart:79-110`). So the success screen still appears before the extrinsic is actually submitted/confirmed. **Mitigation added:** background failures now surface via the `submitFailed` error toast (`transaction_submission_service.dart:198-200` → `multisig_proposal_toast_listener.dart:20`), and the copy is honest ("Transfer proposal submitted" / "Co-signers must approve before the transfer can execute"). The optimistic-then-toast pattern is a defensible design choice, but the core concern (success UI precedes chain confirmation) remains.

**C — Stale fee → Resolved.** `889a0227` removed the `if (_isFetchingFee) return;` early-return and introduced a `_feeFetchGeneration` counter; every change kicks off a fetch and only the latest generation commits its result (`propose_amount_screen.dart:161, 189, 198`). A slower in-flight response for an older amount can no longer overwrite a newer one. (See N2 for a minor follow-up.)

**D — Timeout drops pending state → Not resolved.** `multisig_proposal_polling_service.dart:38-45`: on the 5-minute timeout the poller still just `stopPolling` + `removePendingMultisigProposal` + shows a `timeout` toast, with **no final indexer check / reconcile**. A proposal that actually landed but indexed slowly is silently dropped from the optimistic overlay and the user is told it "timed out". The proposals list providers were invalidated at submit time, so the proposal will reappear on the next full fetch of the multisig screen — data isn't lost forever — but the activity-feed pending row is removed without being swapped for the confirmed creation event, and the toast is misleading. Suggest one final `getProposal`/`searchProposalCreatedByExtrinsicHash` + `reconcileIndexedProposalCreation` before declaring timeout.

---

## New findings (fresh review of the 5 fix commits + general correctness)

- **N1 — (minor, fail-early) Residual zero-fee path.** `_fetchFee` still falls back to `_staticFeeBreakdown(...)` with `networkFee: BigInt.zero` and sets `_hasFee = true` whenever `signer == null` (`propose_amount_screen.dart:181-188`, `151-158`) — the same masking pattern that issue 2 fixed, just on the missing-signer branch. It should be unreachable (the member account is local by construction), but if it ever triggers, the user sees an under-reported fee and an enabled submit button. Treat a missing signer as an error state rather than a zero estimate.
- **N2 — (low, networking) No supersede/cancel on fee fetches.** Removing the in-flight guard (the issue-C fix) means rapid amount edits can launch several overlapping `currentBlockNumber()` + `getFeeForCall()` round-trips; correctness is preserved by the generation check, but superseded requests aren't cancelled/skipped. Consider skipping when a newer generation already exists.
- **N3 — (trivial, DRY) `Constants()` instantiated in 3 places.** After the redundancy/circular-dep fixes, the generated pallet `Constants()` is now created in `MultisigService.palletConstants`, in `MultisigProposal.proposalCreationFeeFor` (a fresh instance per call), and in `PendingMultisigProposalEvent._palletConstants`. The service's two were consolidated; the model/event each kept their own. Could share a single source.
- **N4 — (trivial, DRY) `memberCost` formula duplicated.** `ProposeFeeBreakdown.memberCost` and `PendingMultisigProposalEvent.memberCost` both encode `network/fee + deposit + creationFee`. Acceptable given different field sources, but noted.

### Positive notes
- **GraphQL refactor (`edf1226`) is a real improvement:** open/past/single-proposal queries are now parameterized via GraphQL variables; `_escapeGraphqlString` and string interpolation were removed entirely (no dead code left, lower injection surface).
- **`chain_history_service` fail-early hygiene is good:** `print` → `developer.log`, errors logged with stack traces, and parse failures / unsupported payloads are flagged loudly and only then dropped rather than swallowed silently.
- **The double-proposal fix is the right call** — deferring entirely to `submitExtrinsic`'s internal retries is the correct way to avoid duplicate deposit-reserving proposals.
- **Effective-balance accounting** reuses the shared `_calculatePendingOutgoing` helper for both `balanceProvider` and the new `effectiveBalanceProviderFamily` (good DRY); subtracting the proposer's `memberCost` while pending is reasonable.

---

## Verdict

### Request changes

Most prior feedback is genuinely addressed (1, 2, 3, C, and the redundancy/test nits all resolved cleanly), but a **medium issue remains unresolved (D: timeout drops pending state without reconciliation)**, plus **A is unaddressed** and **B is only partially mitigated** — worth a final pass before merge.

n13 (approved):
Ok A is non-blocking, and fine as it is

Please fix D, it is a good edge case to catch

PR #511, opened 2026-06-09
Execute proposal flow

dewabisma:
## Summary
- Regenerate polkadart metadata
- Implement execute proposal flow
- Update proposal details sheet

---

> [!NOTE]
> **Medium Risk**
> Touches on-chain submission and balance-affecting execute flows with new polling paths; multisig creation preflight no longer reserves deposit, which can change whether creation is allowed vs the old app.
> 
> **Overview**
> Adds an end-to-end **multisig execute** path after approvals hit threshold: confirm sheet (fee estimate + local auth), `TransactionSubmissionService.executeProposal` with optimistic pending state, indexer polling/reconciliation, and UI on proposal detail, activity rows, and error toasts—mirroring the existing approve flow.
> 
> **Proposal detail & lists** now branch on status: **Execute** when `isReadyToExecute`, **Approve** otherwise, with pending “Executing…” states; terminal proposals show an **AT** timestamp from `updated_at` instead of expiry. Open proposals sort by `updated_at`; live proposal resolution checks past proposals too.
> 
> **SDK / chain metadata (spec 129)** is regenerated: multisig **creation deposit** and **dissolve** APIs are dropped from models, preflight, and activity detail; creation cost is pallet + network fees only. `MultisigService` gains build/estimate/submit for `multisig.execute`. GraphQL and `MultisigProposal` use **`updated_at`** for ordering and activity timestamps. Generated pallets also pick up reversible-transfers **guardian** renames and other runtime doc/type updates bundled in the regen.
> 
> **i18n** adds execute-related strings (EN/ID) and tightens insufficient-balance copy for multisig creation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46577df3ea5a8537f0be80c79375c457efb75c41. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (commented):
## Review (agent-assisted, diffed against `multisig-implementation`)

**Verdict: GTG** — no blockers found; two minor, non-blocking notes.

### Minor

1. **Logout doesn't clear pending execution state** — `mobile-app/lib/services/logout_service.dart` clears `pendingTransactionsProvider` and `pendingMultisigCreationsProvider`, but not the new in-memory `pendingMultisigExecutionsProvider` (the same gap already exists for the pending proposals/approvals providers, so this extends a pre-existing pattern). After logout + re-login in the same process, stale "Executing…" rows could survive and execution pollers tied to the old session keep running.

2. **Possible false "executed by another signer" toast** — in `_tryResolveExecutionTimeout` (`multisig_execution_polling_service.dart`), when the proposal status is `executed` but `searchExecutedByExtrinsicHash` returns null, the `executedByOther` toast is always shown. That branch is also reachable when this user's own execution succeeded but account-history indexing is still lagging past the poll timeout — the code comment acknowledges "most likely". Consider softening the copy or rechecking once more before asserting someone else executed it.

### Checked and fine

- `MultisigApprovalToastListener` exhaustively handles all `MultisigExecutionToastKind` values including `executedByOther`.
- Execute submit path (`executeProposal` / `_submitExecuteBackground`) cleans up pending state and surfaces a toast on submit failure; optimistic state + indexer polling mirrors the existing approve flow.

PR #513, opened 2026-06-10
Multisig proposal polling

dewabisma:
## Summary
- Add fetch proposals to polling in multisig account
- Pull to refresh home page also fetch proposals in multisig account

---

> [!NOTE]
> **Low Risk**
> Read-only provider invalidation/refetch scoped to the active multisig account; no changes to signing, auth, or transaction submission.
> 
> **Overview**
> **Multisig proposal data now stays in sync** with the same refresh paths that already update balances and transaction history when a multisig account is selected.
> 
> `polling_refresh_scope` gains `multisigAccountFromActive`, `invalidateActiveMultisigProposals`, and `refreshActiveMultisigProposals` (invalidates open/past proposal providers and current block, then awaits reload). Background global history polls **invalidate** proposals; manual global refresh, home **pull-to-refresh**, silent refresh via the history polling manager, and **account switch** all hook into these helpers (pull-to-refresh and manual global refresh **wait** for fresh proposal data).
> 
> Unit tests cover `multisigAccountFromActive` for multisig vs regular/entrusted/null active accounts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4f175f4d13f7967b60163b33303e84238580210. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (commented):
## Review (agent-assisted, diffed against `beast/cancel-proposal-flow`)

**Verdict: GTG** — small, focused diff; three minor, non-blocking notes.

### Minor

1. **Manual refresh skips reconciliation on proposal-refresh failure** — in `GlobalHistoryPollingService.triggerManualRefresh`, `await refreshActiveMultisigProposals(_ref)` runs before `reconcilePendingTransactions()`. If any proposal/block provider future fails, the method throws and reconciliation never runs for that refresh. The home screen catches the error, but consider try/catching the proposal refresh (or reordering) so reconciliation still happens.

2. **Silent catch in timezone fallback** — `local_notifications_service.dart` has a last-resort `catch (_) {}` with no logging when even the UTC fallback fails. Per repo convention (no silent failures), this should at least `debugPrint`/telemetry the error. Also, the message `'Failed to set device timezone "$e"'` interpolates the exception where the timezone identifier is implied — minor wording nit.

3. **Transient duplicate proposal rows possible** — the background poll now invalidates open proposals (`invalidateActiveMultisigProposals`) independently of pending-event cleanup, so a just-indexed proposal can briefly render alongside its still-pending "Proposing…" row in the Open Proposals section until reconciliation removes the pending event. Cosmetic and self-healing, but worth knowing.

### Checked and fine

- `multisigAccountFromActive` helper + unit tests cover multisig/regular/entrusted/null cases.
- Pull-to-refresh, manual global refresh, silent refresh, and account switch all consistently route through the new `polling_refresh_scope` helpers — good DRY consolidation (also removes the dead `triggerManualRefresh` in `HistoryPollingManager`).
- Reordering polling-service registration before notification init in `AppInitializer` makes listener attachment robust to notification init failures.

PR #515, opened 2026-06-11
Fix double spend: reuse signed extrinsic on retry, remove duplicate retry loop

n13:
## Summary
- `SubstrateService.submitExtrinsic` now signs once and resubmits the same bytes (same nonce) on retry; duplicate-submission errors ("already imported", stale on retry) are treated as success.
- Removed the redundant app-level retry loop in `TransactionSubmissionService`, which re-signed with a fresh nonce per attempt and could double spend.


dewabisma (approved):
LGTM!

PR #518, opened 2026-06-11
Streamlined onboarding: skip phrase screen, backup nudge, blur-first reveal

n13:
## Summary

- **Onboarding**: "Create New Wallet" now creates the wallet directly from the welcome screen — the recovery phrase screen and the caution screen are removed from the create flow (both deleted). Import field obscures pasted words with an eye toggle.
- **Backup nudge**: home screen shows a "Back up your recovery phrase" banner once the wallet holds > 0.0001 QUAN and the phrase has never been exposed. Tapping it opens the existing Settings flow (caution → biometric auth → phrase). The per-wallet flag is set only when the user actually reveals or copies the phrase; importing a wallet sets it immediately. Debug override in `AppConstants.debugAlwaysShowBackupNudge`.
- **Blur-first phrase display**: the single recovery phrase component now always starts blurred with tap-to-reveal/tap-to-hide, and hidden words render a fixed placeholder so word lengths can't be guessed.
- **Misc**: success screen fetches the address checksum on demand (no longer blocks wallet creation), its back button is removed, pull-to-refresh awaits the balance query so balance-driven UI is consistent when the spinner ends, caution-screen checkbox removed (kept for wallet reset), and `[TIMING]` instrumentation added for rpc/indexer queries behind `debugQueryTiming`.

## Test plan

- [x] Create wallet from welcome → lands on "Wallet Created" without seeing the phrase; no back button
- [x] Fund wallet > 0.0001 QUAN → nudge appears (after refresh); faucet button gone
- [x] Nudge → caution → auth → phrase starts blurred; Done without reveal keeps nudge; reveal or copy dismisses it permanently
- [x] Import wallet (paste obscured, eye toggle) → never shows nudge
- [x] Settings → recovery phrase still works; wallet reset still requires checkbox


mobile-app/lib/v2/screens/accounts/account_ready_screen.dart
```diff
@@ -113,11 +104,20 @@ class AccountReadyScreen extends ConsumerWidget {
...
-                            ),
+                            ref
+                                .watch(checksumNameProvider(accountId))
+                                .when(
+                                  data: (checksum) => Text(
+                                    checksum,
+                                    textAlign: TextAlign.center,
+                                    style: text.smallParagraph?.copyWith(
+                                      color: colors.checksum,
+                                      fontWeight: FontWeight.w400,
```
dewabisma, line 115: why set font weight to 400, AFAIK, smallParagraph already 400

dewabisma (approved):
LGTM!

PR #520, opened 2026-06-11
Scanner UX - Safe Area Fix

n13:
Scanner would overlap the iphone camera  on top

Added safe area to prevent this on all devices
IMG_0150


dewabisma (approved):
GTG!

PR #523, opened 2026-06-12
Automate miner release versioning

n13:
## Summary

Replaces the manual version-entry miner release with the same automated flow as quantus-cli / chain / quantus-miner:

- **`create_miner_release_proposal.yml`** (new): dispatch with `patch`/`minor`/`major`/`custom`. Computes the next version from **max(latest `miner-v*` tag, `miner-app/pubspec.yaml`)** — so a manually pre-bumped pubspec can never be silently downgraded — bumps the pubspec, and opens a PR labeled `miner-release-proposal` (labels auto-created on first run). Custom versions below the current version are rejected; an already-existing tag is rejected via a remote `ls-remote` check.
- **`publish_miner_release.yml`** (new): when that PR merges — checks out `merge_commit_sha` directly, verifies pubspec matches the PR-title version, tags the merge commit `miner-vX.Y.Z`, then **builds and releases pinned to that verified merge SHA** (never a bare tag name, which checkout would resolve branch-first, letting a same-named branch shadow the release). The GitHub release is published with `--notes-start-tag` pinned to the previous miner tag so generated notes don't diff against `wallet-v*` tags. The whole run is **idempotent**: re-running after a partial failure skips the existing tag (if it points at the merge commit) and re-uploads assets to the existing release instead of dying. A `concurrency` group serializes publish runs.
- **`create_miner_build.yml`**: now build-only and reusable (`workflow_call` with a `ref` input); manual dispatch still available for test builds. The old in-workflow release job (deprecated `actions/create-release@v1`) is removed.

## Notes

- The proposal uses the `PAT_TOKEN` secret for `gh pr create` so CI triggers on the release PR (same as `ADMIN_PAT` in the other repos). If it has expired, the proposal job fails at PR creation and the secret needs rotating.
- `main` pubspec is at `0.6.0` while the latest tag is `miner-v0.5.0`, so a **patch** dispatch releases `miner-v0.6.1`. To release exactly `0.6.0`, dispatch with **custom** = `0.6.0`.

## Test plan

- [x] actionlint clean on all three miner workflows
- [x] 33 local tests run against the actual scripts extracted from the workflow YAML: version math (incl. pubspec-ahead / tag-ahead / multi-digit `sort -V` cases), custom-version validation and downgrade rejection, pubspec parsing (incl. `+build` suffix and malformed input), PR-title tag extraction, previous-tag lookup with wallet tags interspersed, and tag-push idempotency against a bare origin (fresh, re-run same commit, re-run fresh clone, collision at different commit)
- [x] Full end-to-end run on a disposable repo (`n13/miner-release-e2e-test`) with the two orchestration workflows byte-identical to this branch (pre-SHA-pinning; that later commit only changes which ref the already-proven checkouts receive) and a stub build keeping the exact `workflow_call`/artifact interface:
  - proposal (patch) → PR `ci: Miner version bump to miner-v0.5.1` with correct labels and pubspec bump
  - merge → publish fired, annotated tag created **exactly at the PR merge commit**, build ran at the release commit, release published with all 3 assets and `Full Changelog: miner-v0.5.0...miner-v0.5.1` (wallet tag correctly skipped)
  - full re-run of the publish run → green: "Tag … already points at …; skipping tag creation" + "Release … already exists; re-uploading assets", no duplicate release
  - custom `0.4.0` proposal → rejected in the calculate job (downgrade guard)
  - `is_draft=true` → PR carried `draft-release` label and the release was published as a draft


dewabisma (commented):
Have you tested the CI workflows? It can be tricky to make CI works without actually run and see it works.

n13 (commented):
Verdict: REQUEST_CHANGES — the release build is not pinned unambiguously to the tag/merge commit.

## Blocking finding

1. **[P1] A same-named branch can shadow the release tag and get different code signed and published.** In `.github/workflows/publish_miner_release.yml:88` and `:101`, the workflow passes the bare value `miner-vX.Y.Z` to `actions/checkout`. Checkout v4 resolves an unqualified ref by testing `refs/remotes/origin/<name>` before `refs/tags/<name>`. Therefore, if a branch named `miner-vX.Y.Z` exists, the build jobs check out that branch even though `create-tag` correctly created the tag at the merge commit. The macOS job will then build and Developer-ID-sign the branch contents, and `create-release` will publish those artifacts under the legitimate tag. This bypasses the protected merge/release path for any collaborator able to create a branch. Pass a fully-qualified tag ref (`refs/tags/${{ needs.create-tag.outputs.tag }}`) at both sites, or output the verified merge SHA and use that SHA for every build/release checkout.

The disposable E2E run is real: the two orchestration workflow files are byte-identical to this head, attempt 2 completed successfully, and `miner-v0.5.1` dereferences to the PR merge commit. It did not exercise the branch/tag collision above.

## Non-blocking note

- The PR summary's first-run example is stale after the latest merge from `main`: both base and head now have `miner-app` at `0.6.0` while the latest miner tag is `0.5.0`, so a patch dispatch computes `miner-v0.6.1`, not `miner-v0.5.1`.

## Validation

- Inspected the complete `8e68f1a...4b9888a` diff and intervening fix/merge commits.
- `actionlint v1.7.12` — clean for all three changed workflows.
- `dart run melos bootstrap` — passed for all four packages.
- `dart run melos exec --concurrency=1 -- 'flutter analyze . --fatal-infos'` — all four packages passed.
- `dart run melos exec -- dart format lib test --line-length=120` — passed; 0 files changed.
- `git diff --check` — clean.
- Live GitHub `Analyze` check — passed.

GitHub does not allow an author to submit a native request-changes review on their own PR, so this is posted as a COMMENTED request-changes opinion.

n13 (commented):
Verdict: APPROVE — the prior release-integrity blocker is fixed, and no blocking findings remain.

The new `d904d8a8` commit exports `github.event.pull_request.merge_commit_sha` from `create-tag` and passes that verified SHA to both the reusable miner build and the release checkout (`publish_miner_release.yml:27`, `:91`, `:104`). The tag name is no longer used as a checkout ref anywhere in the publish path, so a same-named branch cannot redirect the code that is built, Developer-ID-signed, or packaged. The PR summary's first-run version example is also now accurate for the current `0.6.0` pubspec / `miner-v0.5.0` tag state.

Validation:

- Inspected the complete `4b9888a..d904d8a` incremental commit and the full resolved `8e68f1a...d904d8a` diff.
- `actionlint v1.7.12` — clean for all three changed workflows, including the new job output/reusable-workflow expression.
- Exact SHA-wiring assertions — passed: one merge-SHA output, two downstream SHA checkouts, and no remaining tag-output checkout.
- `git diff --check` — clean.
- Live GitHub `Analyze` check on `d904d8a` — passed.

No blocking findings remain. GitHub does not allow an author to submit a native approval on their own PR, so this is posted as a COMMENTED approval opinion.

PR #524, opened 2026-06-13
wait for RPC hash before showing success screen

n13:
- send without waiting for server was an over-optimization which led to bad edge cases - if a tx hash is returned, the tx will most likely work 
- show full address in send confirmation - this is a security feature

Simulator Screenshot - iPhone 17e - 2026-06-14 at 11 09 31


dewabisma (approved):
LGTM!

PR #527, opened 2026-06-14
Adding redeem in mobile wallet

n13:
redeem rewards added again
- new system with 7 proofs
- works against new testnet (tested)

### DIFF
--- a/mobile-app/lib/v2/components/address_input_field.dart
+++ b/mobile-app/lib/v2/components/address_input_field.dart
@@ -3,16 +3,21 @@ import 'package:quantus_sdk/quantus_sdk.dart';
 import 'package:resonance_network_wallet/v2/theme/app_colors.dart';
 import 'package:resonance_network_wallet/v2/theme/app_text_styles.dart';
 
-/// Dual-state recipient input. When the controller holds a valid SS58 address
-/// (`hasValid`), shows a pill with the truncated address and optional human
-/// checksum. Otherwise shows a regular search-style input.
+/// Dual-state address input shared by every flow where the user enters a
+/// destination address (send, redeem, ...). When the controller holds a valid
+/// SS58 address (`hasValid`) it shows a pill with the truncated address and the
+/// optional human checksum; otherwise it shows an editable field with an
+/// optional [trailing] action (e.g. a paste button). Tapping the pill clears
+/// the field and refocuses it for re-entry.
 class AddressInputField extends StatelessWidget {
   final TextEditingController controller;
   final FocusNode focusNode;
   final bool hasValid;
   final String? recipientChecksum;
   final String hintText;
-  final bool showSearchIcon;
+  final Widget? trailing;
+  final int addressPrefix;
+  final int addressPostfix;
 
   const AddressInputField({
     super.key,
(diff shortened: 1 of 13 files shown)


mobile-app/lib/v2/components/address_input_field.dart
dewabisma, line 1: I think we should reuse this everywhere we enter address, so we keep it DRY

mobile-app/lib/v2/screens/settings/redeem_address_screen.dart
```diff
@@ -0,0 +1,217 @@
...
+    _recipientController.removeListener(_onRecipientChanged);
+    _recipientController.dispose();
+    _recipientFocus.dispose();
+    super.dispose();
+  }
+
+  Future<void> _prefillPrimaryAccount() async {
+    final settings = ref.read(settingsServiceProvider);
+    final primary = await settings.getAccount(walletIndex: 0, index: 0);
+    if (!mounted || primary == null) return;
```
dewabisma, line 53: Is it necessary checking mounted? we are not setting state

mobile-app/lib/v2/screens/settings/redeem_address_screen.dart
```diff
@@ -0,0 +1,217 @@
...
+  Widget build(BuildContext context) {
+    final colors = context.colors;
+    final text = context.themeText;
+    final fmt = ref.watch(numberFormattingServiceProvider);
+
+    final hasValid = _recipientController.text.trim().isNotEmpty && !_hasAddressError;
+    final amountLabel = fmt.formatBalance(widget.redeemableRewards, maxDecimals: 2, addSymbol: true);
+
+    return ScaffoldBase(
+      appBar: const V2AppBar(title: 'Redeem'),
```
dewabisma: Let's properly localize the copy write for all new text

dewabisma (changes requested):
There is a clarification and two request for changes.

Let's properly keep the localization integrated. I planned to add more locales support as well like the main website.

mobile-app/lib/v2/screens/settings/redeem_address_screen.dart
n13: i guess we're setting the controller text this is why

mobile-app/lib/v2/components/address_input_field.dart
n13: Didn't catch this - you mean to say we already have this? Also I don't like the search thing here either, we're never "searching" for an address ... ???

n13: This feature is only for testnet so lets not over engineer it, lets work on mainnet instead ;)

mobile-app/lib/v2/screens/settings/redeem_address_screen.dart
n13: I guess... I think localization is not the right priority for now, overall though

mobile-app/lib/v2/components/address_input_field.dart
dewabisma: Ohh yeah, we already have input field somewhere.

Hahaha okay, yeah not that much priority but let's not litter the code base at least.

mobile-app/lib/v2/screens/settings/redeem_address_screen.dart
dewabisma: Ohh okay, I thought because we don't cal setState it doesn't need it

mobile-app/lib/v2/screens/settings/redeem_address_screen.dart
dewabisma: Not priority sure, just think since we use AI anyway why not properly do this. Just need to add rules for AI to do the localization.

dewabisma (approved):
LGTM!

PR #529, opened 2026-06-16
All account types

n13:
Refactoring all account types and settings.

## Included
- **Multiple software wallets** — create/import additional wallets alongside the original.
- **Multisig wallets** — create, discover (find multisigs where you're a signer), and propose/approve/cancel/execute. _(behind `enableMultisig` remote flag)_
- **Keystone hardware wallets** _(with `enableKeystoneHardwareWallet` flag)_
  - Scan QR code to add a Keystone account
  - Sign transactions with Keystone
- **Encrypted (wormhole) accounts** — a derived encrypted account per software wallet. _(behind `enableEncryptedAccount` remote flag)_
- **Disconnect accounts & wallets**
  - Disconnect individual accounts (software / keystone / multisig)
  - Removing the last account of an additional wallet deletes the **entire wallet and its recovery phrase from the device** (two-step confirmation)
  - The original wallet (index 0) is protected — extra accounts can be removed, but not the last one (log out instead)
- **Deterministic account indices** — new accounts fill the lowest free derivation index, so removing then re-adding accounts always reproduces the same addresses in the same order (unit-tested).
- Accounts list grouped by wallet + account type; add-account menu reorganized into primary actions + an "Advanced" submenu.

## Notes
- New account types are gated behind remote-config flags for staged rollout (default off).
- Merged latest `main`: BIP-44 gap-limit discovery, global toast provider refactor, silent refresh on notifications, multisig remote flag.
- Added `mobile-app/run_on_device.sh` to build/install/run on iOS 26 devices from the command line (Flutter's Xcode automation is currently broken on iOS 26).

_(Encrypted accounts were previously deferred in this description but are in fact included here, behind a flag.)_


mobile-app/lib/shared/utils/accounts_grouping.dart
```diff
@@ -0,0 +1,146 @@
...
+/// [MultisigAccount.myMemberAccountId]; unresolved ones trail in a standalone
+/// multisig segment. A single software wallet of only transparent accounts with
+/// no multisigs renders flat (no headers).
+AccountsGrouping groupAccounts({required List<Account> accounts, required List<MultisigAccount> multisigs}) {
+  final byWallet = <int, List<Account>>{};
+  for (final a in accounts) {
+    byWallet.putIfAbsent(a.walletIndex, () => []).add(a);
+  }
+
+  final walletIndexByAccountId = {for (final a in accounts) a.accountId: a.walletIndex};
```
dewabisma: wow, first time seeing this way of declaring variable. 

Not sure. if it's good because it's really hard to read by human.

mobile-app/lib/v2/screens/accounts/accounts_sheet.dart
```diff
@@ -162,21 +197,96 @@ class _AccountsScreenState extends ConsumerState<AccountsSheet> {
...
+      AccountSegment.transparent => l10n.accountsSheetSubheaderTransparent,
+      AccountSegment.encrypted => l10n.accountsSheetSubheaderEncrypted,
+      AccountSegment.keystone => l10n.accountsSheetSubheaderKeystone,
+      AccountSegment.multisig => l10n.accountsSheetSubheaderMultisig,
+    };
+    return Padding(
+      padding: const EdgeInsets.only(bottom: 12),
+      child: Text(
+        title,
+        style: context.themeText.smallParagraph?.copyWith(color: context.colors.textTertiary, fontSize: 12),
```
dewabisma: Should have used .details instead of smallParagraph. This way, we don't need to override font size.

mobile-app/lib/v2/screens/accounts/accounts_sheet.dart
```diff
@@ -254,25 +381,27 @@ class _AccountRowShell extends StatelessWidget {
...
+                        if (hasTitle) ...[
+                          Row(
+                            children: [
+                              Flexible(
+                                child: Text(
+                                  title!,
+                                  style: context.themeText.paragraph!.copyWith(
+                                    fontSize: 18,
+                                    fontWeight: FontWeight.w400,
+                                    color: colors.textPrimary,
```
dewabisma: unnecessary color and font weight.

mobile-app/lib/v2/screens/accounts/account_menu_screen.dart
```diff
@@ -24,30 +37,159 @@ class AccountMenuScreen extends ConsumerWidget {
...
+        children: [
+          const SizedBox(height: 8),
+          _ProfileHeader(account: account, colors: colors, text: text),
+          const SizedBox(height: 80),
+          _MenuRow(
+            label: l10n.accountMenuAccountName,
+            value: account.name,
+            onTap: () => _openNameEditor(context, ref, account),
+          ),
+          Divider(color: colors.toasterBackground, height: 1),
```
dewabisma: I think we have many of this repeated,
```
Divider(color: colors.toasterBackground, height: 1)
```

It might be worth it to have reusable divider. I remember I created that before not sure why it's not exist anymore.

mobile-app/lib/v2/screens/send/keystone_scan_signature_screen.dart
```diff
@@ -0,0 +1,276 @@
...
+                  const SizedBox(height: 8),
+                  Text(
+                    l10n.keystoneScanProgress(_seenSeq.length, _expectedParts!),
+                    style: text.detail?.copyWith(color: Colors.white70),
+                  ),
+                ] else if (_parts.isNotEmpty)
+                  Text(l10n.keystoneScanScanning(_parts.length), style: text.detail?.copyWith(color: Colors.white70)),
+                if (AppConstants.debugHardwareWallet && !_submitting) ...[
+                  const SizedBox(height: 16),
+                  TextButton(
```
dewabisma: why do we create new button on the fly here? we must use standard button component

mobile-app/lib/v2/screens/send/keystone_scan_signature_screen.dart
```diff
@@ -0,0 +1,276 @@
...
+            ),
+          ),
+          Positioned(
+            top: 20,
+            left: 16,
+            right: 16,
+            child: SafeArea(
+              child: Row(
+                children: [
+                  IconButton(
```
dewabisma: Same for Icon, we should use quantus icon button.

quantus_sdk/lib/src/services/multisig_service.dart
```diff
@@ -39,9 +39,7 @@ class MultisigService {
   /// Suggested approval threshold at roughly 70% of [signerCount].
   static int defaultThreshold(int signerCount) {
     if (signerCount <= 0) return 1;
-    final threshold = (signerCount * 0.7).round();
-    final minThreshold = signerCount >= 2 ? 2 : 1;
-    return threshold.clamp(minThreshold, signerCount);
+    return (signerCount * 2 / 3).round().clamp(1, signerCount);
```
dewabisma, line 42: Why change this?

This change introduce bad default for when multisig only 2 people.

2 * 2 /3 = 1.3333, if we round it it will be 1. Because clamp now 1, the default will be 1. For 2 people multisig, having signer default to 1 when the signers are 2 people, basically useless.

dewabisma (changes requested):
Okay, found some small parts need to be changed.

quantus_sdk/lib/src/services/multisig_service.dart
n13: No this is what I wanted it to be, it's the formula I want to use, i think I also made tests for it

2: 1/2
3: 2/3
4: 3/4
5: 3/5
6: 4/6
7: 4/7
... etc

its just a default and a 2 people msig is very rate, who knows why they want to do that, could be they share an account (2 signers), or else they want 2/2... it's ok it can be changed by slider

Just the default needs to be reasonable

n13: ok these are the actual expected ones, all good 

 const expected = {1: 1, 2: 1, 3: 2, 4: 3, 5: 3, 6: 4, 7: 5, 8: 5, 9: 6, 10: 7};

dewabisma: Okay, if that is what desired then make sense. Just still weird for 2 people to set signer minimum as 1, like it doesn't prevent anyone from spending it. 

I dunno why you think it's rate maybe because you have seen people use it. But, from what I know in how people use bank account, couple can set need two signers to proceed balance. 

Anyway, I don't really mind being 1 since it can be changed. Not worth fussing it.

dewabisma (approved):
LGTM!

PR #530, opened 2026-06-16
Beast/multisig remote notif

dewabisma:
## Summary
Basically support handling all proposal events notification. 

## Screenshots
Only attached created and executed, but all proposal events tested working.

- Proposal created
Simulator Screenshot - iPhone 8 - 2026-06-16 at 22 24 08

- Proposal executed
Simulator Screenshot - iPhone 8 - 2026-06-16 at 23 20 51

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 7994da76ddf5b73c3f994d6d96018567960a3e9e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (commented):
## Review — multisig remote notifications

Reviewed all 9 files (a1b53fb → bf6c59c). Overall this is a clean, well-structured change with good doc comments and sensible state handling. The cold-start flow (set `proposalIntentProvider` → listen on `multisigAccountsProvider` to retry once loaded → consume intent → `setActiveAccount` → open detail-by-id sheet) is correct and idempotent: the intent is consumed exactly once before the `await`, so there's no double-open or re-entrancy. Routing via `listenManual` on providers (rather than lifecycle hooks / post-frame callbacks) matches the repo's state-machine convention. 

The `debugPrint` → `quantusDebugPrint` migration is a genuine improvement — `quantusDebugPrint` is gated on `kDebugMode`, so the new `FCM tap payload: $data` log won't leak payloads in release builds.

### Confirmed correct
- **Bugbot #1 fix is valid.** `SenotiService.registerDevice` now unions `getAccounts()` + `getMultisigAccounts()`, and `AccountsService.getAccounts()` just delegates to `SettingsService.getAccounts()`, so regular-account registration is unchanged while existing multisigs now register on init/token-refresh.
- `getProposal` returns `null` when not found and throws on HTTP/GraphQL errors, which maps cleanly onto the sheet's `data(null)` → "not found" and `error` → failure message.
- l10n keys (`multisigProposalTitle`, `multisigProposalNotFound`, `multisigLoadFailed`) already exist in both `en` and `id`.

### Non-blocking nits
1. **Dead enum value.** `EventType.PROPOSAL` is added but never referenced anywhere — proposal detection is done by the presence of `multisig`/`proposalId` in `ProposalIntent.tryParse`, not by `type`. Either drop it or wire it in to avoid dead code.
2. **`add()` now blocks on the network.** `MultisigAccountsNotifier.add` does `await _ref.read(firebaseMessagingServiceProvider).insertNewAddress(...)` before updating state, so the UI state update waits on an FCM round-trip. The regular-account paths (`create_account_screen`, `import_wallet_screen`) call `insertNewAddress` fire-and-forget. Consider not awaiting here for consistency and snappier UI (the call already swallows its own errors).
3. **Trailing whitespace** on the blank line after `allAddresses` in `senoti_service.dart` — `dart format` will strip it.
4. **Scope creep:** removing `callRaw` from `MultisigProposal` is unrelated to push notifications. It's harmless (the GraphQL query still selects `call_raw`; it's just ignored now, and no UI reads `callRaw`), but it muddies the diff.

### Acknowledged
- Bugbot #2 (foreground proposal pushes don't render an in-app notification, since `_remoteMessageToNotificationData` only handles tx-shaped payloads) — fine to defer until there's a local notification surface, as discussed in the thread.

### Verdict
**Approve** once nit #1 (dead `EventType.PROPOSAL`) and #3 (formatting) are addressed — both trivial. Nits #2 and #4 are optional. Functionality looks correct per the static review and the attached screenshots.

n13 (approved):
Not sure it's worth removing the dead enum item - approved anyway

n13 (approved):
## Re-review — multisig remote notifications

Re-reviewed at HEAD `51e075c`. The only change since my last pass is the trailing-whitespace fix in `senoti_service.dart` (nit #3) — done. Re-verified the functional flow end to end:

- **Cold-start routing is correct and idempotent.** `navigateToProposalFromPayloadIfPossible` sets `proposalIntentProvider` from the FCM payload (and returns `true` to skip the tx path); `_onProposalIntent` consumes the intent exactly once (sets it null before the `await`), retries via the `multisigAccountsProvider` listener while accounts are still loading, sets the active account, then opens the sheet. `_MultisigProposalDetailSheetById` renders loader / `not found` / `load failed` off `multisigProposalByIdProvider`, and pops an existing detail sheet first so there's no double-open.
- **Bugbot #1 fix holds.** `registerDevice` unions `getAccounts()` + `getMultisigAccounts()`, so existing multisigs register on init / token-refresh.
- `ProposalIntent.tryParse` correctly handles FCM `proposalId` as string or number; l10n keys (`multisigProposalTitle`, `multisigProposalNotFound`, `multisigLoadFailed`) exist in both `en` and `id`.
- `debugPrint` → `quantusDebugPrint` (gated on `kDebugMode`) keeps payloads out of release logs.

Remaining items are unchanged and already acknowledged: dead `EventType.PROPOSAL` (you opted to keep it), `MultisigAccountsNotifier.add` awaiting `insertNewAddress` (optional), and Bugbot #2 — foreground proposal pushes don't render an in-app notification, deferred until there's a local-notification surface.

### Verdict
**Approve** (re-affirmed).

PR #532, opened 2026-06-19
Implement BIP-44 gap-limit account discovery

n13:
## Summary
- Replace the single fixed batch of 20 HD indices with a proper BIP-44 gap-limit scan: keep scanning in batches until `gapLimit` (20) consecutive indices have no on-chain account.
- Remove the obsolete legacy raw account (`index -1`) check; rewards now go to a wormhole account, so no `index 0` rewards account exists anymore. This also drops the now-unused `SubstrateService` dependency.

## Test plan
- [ ] Import a wallet with >20 used HD accounts and confirm all are discovered.
- [ ] Import a wallet with sparse usage (gaps) and confirm discovery stops after 20 consecutive empty indices.
- [ ] Import a fresh mnemonic and confirm no extra accounts are added.


dewabisma (approved):
LGTM! Now it works as it should be!

PR #534, opened 2026-06-19
feat: simplify toast provider

dewabisma:
## Summary
Instead of having many provider for toast, we have a global toast provider now. 

---

> [!NOTE]
> **Low Risk**
> Refactor-only UI notification wiring with no change to submission or polling business logic; main risk is a missed call site leaving a toast unsurfaced.
> 
> **Overview**
> Replaces five multisig-specific toast providers and three nested listener widgets with a single **`globalToastProvider`** and **`GlobalToastListener`**.
> 
> Producers (multisig polling and submission services) now push **already-localized** strings via `showSuccess` / `showError` / `showInfo` instead of emitting typed enum events. **`HomeScreen`** wraps content in one listener instead of stacking creation, approval, and proposal listeners.
> 
> User-visible multisig toast copy and styling paths are unchanged; only the plumbing is unified so any `Ref` can surface toasts without `BuildContext`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 780f846dbea434a776a89f7d453dd517325100e6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (approved):
## Review: simplify toast provider

Solid, well-scoped simplification. Five multisig-specific toast providers + three nested listener widgets collapse into a single `globalToastProvider` (`GlobalToastNotifier`) and one `GlobalToastListener`. I verified all 14 call sites and every one preserves the original severity/toaster type and the exact l10n string, so user-visible behavior is unchanged. Localization correctly moves to producers via `ref.read(l10nProvider)` (a context-free `Provider`, safe from a service `Ref`). No dangling imports/references to the deleted files remain — it compiles. Net result is a strong DRY win (8 files → 2).

### Findings

- **[Minor] "Queue" is actually a single nullable slot.** `globalToastProvider` holds one `ToastMessage?`, but the doc comment calls it an "App-wide toast queue." Toasts emitted within the same microtask coalesce — Riverpod delivers only the last transition, so the earlier one is dropped. Previously, separate providers meant cross-category toasts (e.g. a creation-ready + an approval-timeout firing together) could each display; they now share one slot. Real-world risk is low since each event is usually its own async callback, but please either implement a real queue or soften the doc wording to avoid implying a guarantee that isn't there.

- **[Minor] Stale comment referencing deleted symbols.** `mobile-app/lib/v2/screens/multisig/add_multisig_screen.dart` (lines ~272–274) still says "surface a toast via the creation toast listener (`MultisigCreationToastKind.submitFailed`)" — both the listener and that enum are deleted in this PR. Update it to reference the global toast flow.

- **[Nit] Unused API surface.** `showWarning` / `showCopy` and `ToastType.warning` / `copy` have no call sites. They compile (matching `showWarningToaster` / `showCopyToaster` extensions exist), but per the repo's "concise/minimal" convention consider dropping them until a consumer needs them.

- **[Nit] No tests.** No unit test for the new central `GlobalToastNotifier`/listener. Consistent with the repo (no existing toast tests), so non-blocking — a tiny notifier test would be cheap insurance for this now-shared path.

### Standards check
- Fail-early: preserved — catch blocks still log (`quantusDebugPrint` + telemetry) and `rethrow`; toasts remain supplementary. No new silent failures.
- DRY: clear improvement, the main goal of the PR.

> Note: this PR is **stacked on #533** (`beast/optimize-multisig-tx-submission`) — its diff is correct only against that base, and it should be merged **after** #533.

**Verdict:** APPROVE (non-blocking minor/nit comments only)

n13 (approved):

PR #539, opened 2026-06-23
Fix the button layout on receive

n13:
Adds the copy button back - it was flipped vs the design, which made me complain about it to the designer

All buttons tested ✅

# Screens

Simulator Screenshot - iPhone 17e - 2026-06-23 at 16 11 48

Simulator Screenshot - iPhone 17e - 2026-06-23 at 16 12 01

# Design

image


dewabisma (approved):
LGTM!

PR #544, opened 2026-06-30
Scope backup nudge to wallets created on-device

n13:
## Summary
- Only show the recovery-phrase backup nudge for mnemonic-backed wallets **created on this device** (or legacy wallets of unknown origin, for backwards compatibility).
- Never nudge **keystone/hardware** wallets, **multisig/entrusted** accounts, or **imported** wallets (you already have the phrase when importing).
- Track wallet origin (`created` vs `imported`) via a new `WalletOrigin`, persisted per `walletIndex` in `SettingsService`.
- Add an **"I already backed up my wallet"** action on the recovery-phrase caution screen (shown from the home nudge) so users who already backed up can dismiss without viewing/copying the phrase again.

## Test plan
- [ ] Create a new wallet on-device, fund it above the threshold → nudge appears on home.
- [ ] Tap nudge → "I already backed up my wallet" → nudge disappears (no phrase reveal needed).
- [ ] View/copy the phrase → nudge also disappears (existing behavior).
- [ ] Import a wallet → no nudge.
- [ ] Keystone (hardware) wallet active → no nudge.
- [ ] Multisig/entrusted account active → no nudge.
- [ ] Legacy wallet (no stored origin) still nudges.

### DIFF
--- a/mobile-app/lib/v2/screens/settings/recovery_phrase_confirmation_screen.dart
+++ b/mobile-app/lib/v2/screens/settings/recovery_phrase_confirmation_screen.dart
@@ -1,15 +1,20 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:resonance_network_wallet/providers/l10n_provider.dart';
+import 'package:resonance_network_wallet/providers/wallet_providers.dart';
 import 'package:resonance_network_wallet/services/local_auth_service.dart';
 import 'package:resonance_network_wallet/shared/extensions/toaster_extensions.dart';
 import 'package:resonance_network_wallet/v2/screens/settings/recovery_phrase_screen.dart';
 import 'package:resonance_network_wallet/v2/screens/settings/settings_caution_scaffold.dart';
 
 class RecoveryPhraseConfirmationScreen extends ConsumerStatefulWidget {
-  const RecoveryPhraseConfirmationScreen({super.key, required this.walletIndex});
+  const RecoveryPhraseConfirmationScreen({super.key, required this.walletIndex, this.showAlreadyBackedUp = false});
 
   final int walletIndex;
+
+  /// When true (the home backup nudge), offers an "I already backed up" action
+  /// that dismisses the nudge without revealing the phrase.
+  final bool showAlreadyBackedUp;
 
   @override
   ConsumerState<RecoveryPhraseConfirmationScreen> createState() => _RecoveryPhraseConfirmationScreenState();
@@ -31,6 +36,12 @@ class _RecoveryPhraseConfirmationScreenState extends ConsumerState<RecoveryPhras
     }
   }
 
+  void _onAlreadyBackedUp() {
+    ref.read(settingsServiceProvider).setRecoveryPhraseViewed(widget.walletIndex);
+    ref.invalidate(recoveryPhraseViewedProvider(widget.walletIndex));
+    Navigator.of(context).pop();
+  }
+
   @override
   Widget build(BuildContext context) {
     final l10n = ref.watch(l10nProvider);
@@ -40,6 +51,8 @@ class _RecoveryPhraseConfirmationScreenState extends ConsumerState<RecoveryPhras
       data: SettingsCautionScaffoldData.recoveryPhrase(l10n),
       continueLabel: l10n.commonContinue,
       onContinue: _onContinue,
+      secondaryLabel: widget.showAlreadyBackedUp ? l10n.settingsRecoveryAlreadyBackedUp : null,
+      onSecondary: widget.showAlreadyBackedUp ? _onAlreadyBackedUp : null,
     );
   }
 }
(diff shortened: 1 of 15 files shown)


n13 (commented):
## Review: Scope backup nudge to wallets created on-device (#544)

### Summary of changes
The PR scopes the home backup/recovery-phrase nudge to only on-device created (or legacy unknown-origin) mnemonic wallets. It introduces `WalletOrigin` tracking, sets it on create/import paths, updates the `backupReminderWalletIndexProvider` with the necessary guards, and adds an "I already backed up my wallet" secondary action (only surfaced from the home nudge) to dismiss without forcing phrase reveal.

### What looks good
- **Correct scoping logic**: `RegularAccount` + `accountType == local` + `origin != imported` + `!viewed` + balance threshold. Hardware, multisig, entrusted, and imported wallets are correctly excluded.
- **Legacy compatibility**: `getWalletOrigin` returning `null` (unknown) still allows the nudge, matching the stated intent.
- **UX improvement**: The new "I already backed up" path from the banner is a nice touch and only wired when `showAlreadyBackedUp: true`.
- **Settings storage**: Origin is stored per `walletIndex` (same granularity as viewed/mnemonic), using the existing prefs pattern.
- **No linter/analyze issues** on changed files; the "Analyze" CI check is green.
- Translations added for both `en` and `id` (plus generated localizations).

### Minor issues / suggestions (non-blocking)
1. **Test coverage gap**  
   `wallet_creation_service_test.dart` verifies mnemonic + account + referral, but does not assert the new `setWalletOrigin(WalletOrigin.created)` call (in the `!hasRoot` branch) or `verifyNever` in the has-root branch.  
   The import path that now calls `setWalletOrigin(imported)` also lacks a corresponding assertion/test. Recommend adding:
   ```dart
   verify(settings.setWalletOrigin(0, WalletOrigin.created)).called(1);
   ```
   (and the negative case).

2. **Slight duplication in wallet bootstrap paths (DRY)**  
   Root account creation + mnemonic persist lives in `WalletCreationService` (used by welcome) and is duplicated inline in `import_wallet_screen.dart` (setMnemonic + addAccount + discover). Origin is now set in both.  
   Not introduced by this PR, but a good future refactor would be to add an `importWallet(...)` helper in the service so both creation and import go through a single place for side effects like origin.

3. **Cleanup on wallet removal (hygiene)**  
   `settings_service.removeWallet` deletes the mnemonic but does not remove the `wallet_origin_$n` or `recovery_phrase_viewed_$n` keys. Since `nextWalletIndex` can reuse freed indices, a subsequent create/import on a reused index will overwrite, so no correctness bug today. Still, for cleanliness consider:
   ```dart
   _prefs.remove(_walletOriginKey(walletIndex));
   _prefs.remove(_recoveryPhraseViewedKey(walletIndex));
   ```
   (out of scope for this change).

4. **Minor inconsistency**  
   Import explicitly invalidates `walletOriginProvider` after setting; the welcome/create path does not. Because the provider is a plain derived `Provider` that reads prefs on access, it is not required for correctness, but adding the invalidate in `welcome_screen.dart` would be consistent.

5. **Debug flag interaction**  
   `debugAlwaysShowBackupNudge` is evaluated after the Regular+local type checks (good), so it won't force the nudge on hardware/imported in debug either.

### Nitpicks
- The provider now has a multi-line doc comment explaining the rules. If the project prefers very minimal comments, the doc could be trimmed, but the explanation is helpful for maintainers.
- No behavior changes to multisig/entrusted/hardware flows or serialization.

### Verdict
**Approve**

The change is focused, safe, and solves the stated problem without side effects on other account types. Ready to merge (test coverage improvement would be nice to have).

mobile-app/lib/v2/screens/settings/recovery_phrase_confirmation_screen.dart
```diff
@@ -1,16 +1,21 @@
...
 
 class RecoveryPhraseConfirmationScreen extends ConsumerStatefulWidget {
-  const RecoveryPhraseConfirmationScreen({super.key, required this.walletIndex});
+  const RecoveryPhraseConfirmationScreen({super.key, required this.walletIndex, this.showAlreadyBackedUp = false});
 
   final int walletIndex;
 
+  /// When true (the home backup nudge), offers an "I already backed up" action
+  /// that dismisses the nudge without revealing the phrase.
+  final bool showAlreadyBackedUp;
```
dewabisma, line 17: is this necessary? we can just not displaying it at all if user already back up, can't we?

dewabisma (commented):
Overall looking good, just one thing to clarify.

dewabisma (approved):
Actually will approve, but need clarification on the intent.

mobile-app/lib/v2/screens/settings/recovery_phrase_confirmation_screen.dart
n13: It's an option on the screen because the screen is also used from settings. So when users open it from settings the button doesn't make sense.

dewabisma: Ahh okayy

PR #549, opened 2026-07-02
V12 SDK Fixes

illuzen:
## Summary

Security review remediation for quantus_sdk addressing 20 findings from the V12 security audit.

### Critical/High Severity
- **#93270** - Multisig approvals now verify `call_raw` matches indexer-provided recipient/amount before signing
- **#93266** - Account signing validates account type and derived address matches accountId
- **#93267** - Dev mnemonics blocked in release builds

### Medium Severity
- **#93172** - Prevent SSRF via path concatenation host override in `RedundantEndpointService`
- **#93180** - Bound recent address storage (max 256 chars) to prevent DoS
- **#93189** - Fix recovery proxy lookup double-hashing account ID
- **#93193** - Remove hard-coded test address from `getInterceptedAccounts`
- **#93199** - Exclude wormhole accounts from Senoti registration to preserve privacy
- **#93263** - Validate SS58 prefix matches expected network
- **#93264** - Improve RPC failover robustness with exponential backoff and 5xx detection
- **#93265** - Reject ambiguous locale separators in wire amount parsing
- **#93272** - Use correct pallet for guardian fund recovery
- **#93282** - Prevent stale async init from mutating cleared/disposed state
- **#93096** - Validate exchange rate data (ISO 4217 keys, rate bounds, expiry cap)
- **#93184** - Enforce maxSigners=100 limit in multisig event parsers
- **#93198** - Reject negative fee values in MultisigCreationEvent
- **#93276** - Fix paste detection in DecimalInputFilter

### Other Fixes
- Add request timeout (30s default) to `RedundantEndpointService`
- Bound account discovery (gapLimit 1-100, maxScanIndex 10000)
- Add mutex to prevent account persistence race conditions
- Fix pagination cursor drift in ChainHistoryService
- Handle block-number delays in HighSecurityService
- Bound address length and cache size in HumanReadableChecksumService
- Fix migration to use each account's walletIndex and handle encrypted accounts
- Correct pallet indices in payload parser documentation (#93274, #93275)

## Testing

All fixes include unit tests verifying the security boundary. Run with:
```bash
cd quantus_sdk && flutter test

Notes
Each fix uses minimal code changes rather than wholesale refactoring
Ambiguous wire amounts (e.g., "1,000" or "1.000") now return null rather than guessing
Encrypted accounts throw UnsupportedAccountTypeForSigningException from getKeypair() since they use ZK proofs

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **High Risk**
> Touches authentication-adjacent signing (`getKeypair`), multisig approval trust, migration data loss paths, and network request routing—errors could block legitimate use or, if incomplete, leave spoofing vectors open.
> 
> **Overview**
> Implements **V12 security audit** fixes across `quantus_sdk`, mobile wallet, and cold-wallet signing UI—mostly hardening validation, failover, and user-facing safety around signing and migration.
> 
> **Multisig & signing trust** — Multisig proposals now decode indexer `call_raw` and compare recipient/amount to displayed fields (`CallVerificationStatus`, `isVerified` / `hasVerificationMismatch`). Multisig indexer rows are skipped when predicted on-chain address does not match. `Account.getKeypair()` rejects keystone/external/encrypted types and throws if derived SS58 ≠ stored `accountId`. Dev crystal mnemonics are blocked when `allowDevMnemonics` is false.
> 
> **Network & parsing** — `RedundantEndpointService` adds per-request timeouts, failure cooldown/backoff, 5xx failover, and safe URI building to block SSRF via malicious paths. Exchange rates are filtered (ISO 4217 keys, bounded rates, capped expiry). Wire amount parsing returns null on ambiguous single-separator inputs; `DecimalInputFilter` distinguishes true single-char typing from paste/replace and validates grouping via `normalize()`.
> 
> **Migration & high security** — Migration returns per-account `MigrationSuccess` / `MigrationFailure` (per-wallet mnemonic, encrypted/wormhole derivation); mobile UI shows success/failure counts, disables migrate when none succeed, uploads only successes, and avoids clearing old accounts if any fail. Guardian recovery uses `reversibleTransfers.recoverFunds` instead of a recovery utility batch; high-security delay is modeled as timestamp or block delays (`SafeguardDelay`). Reversible payload display uses `ReversibleDelay` with correct units (cold wallet sign screen updated).
> 
> **Other bounds** — Account discovery caps gap limit and max scan index and ignores extra indexer IDs; history pagination advances by raw rows consumed; multisig signers capped and fees must be non-negative; checksum service LRU-bounds cache and address length; connectivity service guards disposed state.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1a5b0bb8fec1ba021a1af6091f56c4d0086cafc. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


n13 (changes requested):
#93270 - Multisig approvals now verify call_raw matches indexer-provided recipient/amount before signing
#93266 - Account signing validates account type and derived address matches accountId
#93267 - Dev mnemonics blocked in release builds

None of these are real bugs

I don't think it makes sense to do this

The entire app and all flows have to be re-tested after this. Why were 4,000 lines of code added? What do they do?

n13 (changes requested):
## Review

**Verdict: Request changes.**

This PR bundles a handful of genuinely valuable bug fixes with a large amount of defensive hardening whose security impact is overstated, ~600 lines of generated-binding churn, and a red CI. The real fixes are worth landing, but the PR needs trimming and the blocking issues below fixed first.

### The security framing is overstated

**#93270 (call_raw verification) — the headline critical — does not hold up:**

- `call_raw`, `transfer_to`, and `transfer_amount` all come from the **same indexer record**. A compromised indexer serves a self-consistent lie (fake `call_raw` matching the fake display fields) and verification passes. This is a self-consistency check within a single untrusted source, not a trust boundary. It only catches an attacker who tampers the display columns but forgets `call_raw`.
- **Nothing consumes it.** `isVerified` / `hasVerificationMismatch` are referenced nowhere outside the model and its tests — no approve/execute UI checks them. As merged, this changes no behavior at all.
- The approve extrinsic commits only to `proposalId` (the PR's own test demonstrates this), so the meaningful fix is to verify the proposal against **on-chain storage via RPC** before signing, and have the approval UI block on that. Either do that, or don't call this a critical fix.

By contrast, the `discoverForUser` check (skip records whose address ≠ locally computed `hash(signers, threshold, nonce)`) **is** sound — the address cryptographically commits to the signer set. Good fix, but see blocking issue 6.

Similar pattern elsewhere: the SSRF fix guards paths that no call site passes untrusted data into (all callers use `post(body:)` with fixed/empty paths); the exchange-rate bounds still accept wrong-but-plausible rates from a compromised rate API; the DoS bounds (address length, LRU cache, gap limit) are local-resource hygiene. Fine to have, but low severity.

**The fixes actually worth landing** are mostly correctness, not security — and they're good:
- Migration fix (wallet-index/encrypted derivation): the old code derived accounts from the wrong seed and then **cleared the old data** — permanent fund loss. Most impactful fix in the PR.
- #93272 guardian recovery used a pallet that was never configured — the rescue path could not work.
- #93189 recovery proxy double-hashing (wrong address returned).
- #93193 leftover hard-coded test address.
- `clearAll()` no longer silently deleting mnemonics; pagination cursor drift; #93199 Senoti wormhole privacy; #93263 SS58 prefix validation (panic→error is also a good change).

### Blocking issues

1. **CI Analyze is red** (`melos exec flutter analyze`, warnings are fatal). mobile-app: 3× `invalid_return_type_for_catch_error`, 1× unused `_hasMigrationFailures`. quantus_sdk: unused `wormholeKeyPair` local, unused imports/locals in new tests.
2. **`RecentAddressesService.addAddress` return type change (`Future<void>` → `Future<bool>`) breaks the three `.catchError((e) => debugPrint(...))` call sites** — the handler now must return `FutureOr<bool>`; if the future ever fails, the handler itself throws a runtime `TypeError` inside `unawaited(...)`. Fix the call sites or keep the void return.
3. **`Account.getKeypair()` encrypted branch is dead code**: it derives the wormhole keypair, then unconditionally throws. The exception message also claims encrypted accounts *can* derive keypairs, contradicting the behavior. Delete the derivation and fix the message/doc comment.
4. `_hasMigrationFailures` is written but never read — either surface it in UI or drop it.
5. **Wire-amount ambiguity check is over-broad**: `parseWireAmount('1234.567')` returns null, but a grouping reading is impossible there (first group would be 4 digits). Only treat single-separator + 3 trailing digits as ambiguous when the integer part is 1–3 digits. As written, this will reject legitimate payment-URL/QR amounts.
6. **`discoverForUser` silently `continue`s** on mismatch/parse failure with no logging — a legit multisig disappearing from a user's list with zero diagnostics will be unsupportable. Also, when the indexer omits `nonce`, the code assumes `defaultMultisigNonce`; a legit multisig created with a different nonce would be silently hidden. Log every skip with the reason.
7. **`_buildSafeUri` changes URL semantics**: empty path now produces a trailing slash (`…/v1/graphql/` — I verified the production Hasura tolerates this, 200), and a leading-`/` path now **replaces** the base path instead of appending (old: `$url$path`). Latent breakage for any future endpoint with a path prefix; the double-negative validation logic should also be simplified — it's hard to convince yourself it's airtight.
8. `TransactionInfo`'s deprecated `reversibleTimeframe` constructor parameter is accepted and **silently dropped**. Remove it or make it populate `reversibleDelay`.
9. **Diff noise**: `frb_generated.*` reformatting (~600 lines) and `pubspec.lock` downgrades (meta 1.18.0→1.17.0, test_api 0.7.11→0.7.10) suggest the bindings were regenerated with a different Flutter/FRB toolchain than the repo's. Regenerate with the pinned toolchain so the diff shows only the real change (the `ss58_to_account_id` error codec).
10. Minor: PR description says dev mnemonics are "blocked in release builds", but `allowDevMnemonics` defaults to `AppConstants.globalDebug`, which is `const false` — they're blocked everywhere unless code is edited. Fine as behavior, but say what it does.

### Verification performed

- `flutter test` in quantus_sdk: **220 passed**; 3 test files (`account_extension_test`, `generate_keys_test`, `recovery_proxy_encoding_test`) require the native Rust dylib and can't run from a clean checkout — `cargo build` fails to resolve the yanked `core2 ^0.3` (via KeystoneHQ `libflate`) because `rust/Cargo.lock` isn't committed. Worth fixing separately; note CI isn't running these tests either.
- `flutter analyze`: cold-wallet-app clean; mobile-app and quantus_sdk warnings as listed above (matches the failing CI job).
- Rust changes reviewed: multisig canonicalization/threshold validation and SS58 prefix validation look correct.

### Recommendation

Split out and land the genuine bug fixes (migration, guardian recovery pallet, proxy double-hash, test-address removal, clearAll, pagination) in a slim PR. For #93270, either make verification meaningful — check against on-chain proposal state and wire `isVerified`/`hasVerificationMismatch` into the approve/execute UI — or drop the 200 lines of unused machinery. Revert the generated-file and lockfile churn.

PR #550, opened 2026-07-02
Restore quantus_sdk/rust Cargo.lock to unbreak Rust builds

n13:
## Summary

Fresh checkouts of this repo currently cannot build the Rust SDK crate. `cargo build` in `quantus_sdk/rust` fails during dependency resolution:

```
error: failed to select a version for the requirement `core2 = "^0.3"`
candidate versions found which didn't match: 0.4.0, 0.3.3, 0.3.2, ...
```

**Root cause:** `ur-registry` 0.0.52 (pulled in via `quantus_ur` 1.1.0 → KeystoneHQ/keystone-sdk-rust tag 0.0.52) depends on `core2 ^0.3`, and *every* version of `core2` on crates.io has since been yanked. Fresh resolution refuses yanked versions, so any clean clone / new machine / cleared cargo cache fails to build.

This went unnoticed because `Cargo.lock` (which pinned `core2 0.3.3`) was tracked until #542 deleted it three days ago ("uncommit generated and sensitive files") and added a blanket `*.lock` gitignore. Yanked crates remain downloadable when pinned in a lockfile, so restoring the lock fixes the build.

## Changes

- Restore `quantus_sdk/rust/Cargo.lock` exactly as it was before #542 (the manifest hasn't changed since, so the lock is still consistent)
- Add `!Cargo.lock` to `.gitignore` so it stays tracked

`Cargo.lock` should be committed anyway — this crate builds a binary artifact (the FRB dylib), and Cargo's own guidance is to commit lockfiles for reproducible builds. It's not a generated-noise file like FRB bindings; it's the only thing guaranteeing every dev and CI machine compiles the same dependency tree.

## The longer-term fix

KeystoneHQ has already solved this upstream: `ur-registry` is now published on crates.io, and **1.0.3+ drops `core2` entirely** (uses upstream `libflate 2.2.2` + `no_std_io2`; latest is 1.0.6, June 2026). However, `quantus_ur` — including its latest tag 1.3.0 and main — still pins the old git tag 0.0.52, and 0.0.x → 1.0.x is a breaking bump with API changes. Migrating means updating the `quantus_ur` repo first, then bumping the pin here. Filed as follow-up work; this PR unblocks builds today.

## Test plan

Verified in a clean worktree from `main` (no pre-existing `target/`):

- [x] `cargo build --release --locked` succeeds (previously failed at resolution)
- [x] `cargo test --release --locked` — 13/13 pass
- [x] `flutter test` in `quantus_sdk` with the freshly built dylib — 85/85 pass (including the 3 native-dependent test files that were unrunnable before)


n13 (commented):
Reviewed with independent verification of each claim:

**Verified**
- The restored `quantus_sdk/rust/Cargo.lock` is byte-identical to the pre-deletion version (blob `4ba58c8` in both `441b7de2^` and this PR).
- `quantus_sdk/rust/Cargo.toml` is unchanged since the deletion, and `cargo metadata --locked` succeeds on this branch, so the lockfile is consistent with the manifest.
- Confirmed via crates.io API that every published version of `core2` (0.0.0 through 0.4.0) is yanked, so fresh resolution of `ur-registry 0.0.52`'s `core2 ^0.3` requirement genuinely fails — restoring the lock is the correct immediate fix.
- Branch is up to date with `main`.

**One correction to the PR description**
#542 did not add the blanket `*.lock` gitignore — that rule has been in `.gitignore` since the initial commit (Mar 2025). The lockfile stayed tracked despite it because gitignore doesn't affect already-tracked files; #542 just `git rm`'d it. The conclusion is unchanged: the `!Cargo.lock` negation added here is required, otherwise the restored file would be the only thing keeping itself tracked and would silently disappear again if ever deleted.

**Minor side effect worth knowing**
`!Cargo.lock` un-ignores every `Cargo.lock` in the repo, so `rust-transaction-parser/Cargo.lock` (previously ignored, never tracked) now shows up as untracked for anyone who has built that crate. Since it also builds a binary, the same commit-the-lockfile reasoning applies — consider committing it too, here or in a follow-up.

LGTM otherwise — fix is minimal, correct, and the test plan checks out.

dewabisma (approved):
LGTM!

PR #551, opened 2026-07-02
Bump quantus_ur to 1.4.0 (drops yanked core2)

n13:
Bumps `quantus_ur` 1.1.0 → 1.4.0 (Quantus-Network/quantus_ur#1), which switches to Keystone's crates.io releases and removes the yanked `core2` crate from the dependency tree.

Verified: `cargo build --locked`, 13/13 Rust tests, 85/85 SDK Flutter tests.


dewabisma (approved):
Great!

PR #552, opened 2026-07-02
fix: correctness fixes from v12 review

n13:
Lands only the fixes worth keeping from #549 (see review there), rebuilt on main:

- **Migration** (most impactful): derive each account from its own wallet mnemonic and account type (wormhole derivation for encrypted accounts); old accounts are only cleared when every migration succeeded — the old code derived from the wrong seed and cleared old data anyway, causing permanent fund loss.
- **#93272 guardian recovery**: use `reversibleTransfers.recoverFunds` instead of batching calls to the recovery pallet, which was never configured, so the rescue path could not work.
- **#93189 recovery proxy**: encode the storage-returned AccountId32 directly; `toAccountId` was double-hashing and returning the wrong address.
- **#93193**: remove leftover hard-coded test address from intercepted accounts.
- **Pagination**: advance history cursors by raw rows consumed, not parsed count, to prevent cursor drift.
- **#93199 Senoti privacy**: wormhole accounts are excluded from device registration (and from the Supabase migration upload).

**Intentionally not fixed**:
- **clearAll()**: no longer silently deletes mnemonics; full logout uses explicit `clearAllIncludingMnemonics()`.
- **#93263 SS58 validation**: `ss58ToAccountId` validates the network prefix and returns an error instead of panicking.

clearAll change was simply worse (and more) code
ss58 check is not needed and was more code, and added some risks of failures we would now have to check for.

### DIFF
--- a/quantus_sdk/lib/src/services/senoti_service.dart
+++ b/quantus_sdk/lib/src/services/senoti_service.dart
@@ -59,15 +59,18 @@ class SenotiService {
   final SettingsService _settingsService = SettingsService();
   SenotiAuthClient get _client => SenotiAuthClient(AppConstants.senotiEndpoint);
 
+  /// Wormhole addresses are meant to be unlinkable to the user's identity, so
+  /// registering them with the notification service would deanonymize them.
+  static List<String> notifiableAddresses(List<Account> accounts, List<MultisigAccount> multisigAccounts) => [
+    ...accounts.where((a) => a.accountType != AccountType.encrypted).map((a) => a.accountId),
+    ...multisigAccounts.map((a) => a.accountId),
+  ];
+
   Future<void> registerDevice(String token, String platform) async {
-    // Wormhole addresses are meant to be unlinkable to the user's identity, so
-    // registering them with the notification service would deanonymize them.
-    final regularAddresses = (await _settingsService.getAccounts())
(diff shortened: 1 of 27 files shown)


quantus_sdk/lib/src/services/senoti_service.dart
```diff
@@ -59,10 +59,18 @@ class SenotiService {
   final SettingsService _settingsService = SettingsService();
   SenotiAuthClient get _client => SenotiAuthClient(AppConstants.senotiEndpoint);
 
+  /// Wormhole addresses are meant to be unlinkable to the user's identity, so
+  /// registering them with the notification service would deanonymize them.
+  static List<String> notifiableAddresses(List<Account> accounts, List<MultisigAccount> multisigAccounts) => [
```
dewabisma, line 64: Do we really want to not notify them? I mean if they receive fund, isn't it better to notify them? Or we really don't want to register the address at all?

dewabisma (commented):
LGTM,  just have one question to consider. I will approve once confirmed.

quantus_sdk/lib/src/services/senoti_service.dart
n13: Can't notify them yes - otherwise we just give away everyone's secret addresses lol

dewabisma (approved):

PR #555, opened 2026-07-02
Fix parser criticals: parse full signed payload, validate network

n13:
## Summary

Ports the hardware-wallet parser hardening from [keystone3-firmware PR #6](https://github.com/Quantus-Network/keystone3-firmware/pull/6) (`rust/apps/quantus/src/parser.rs`, two Critical audit findings) to this repo.

### `rust-transaction-parser` (near 1:1 mirror of the firmware parser)
- Replaces hand-sequenced field reads with a declarative `#[derive(Decode)]` mirror of the runtime call enums; `#[codec(index)]` matches the current runtime (Balances = 2, ReversibleTransfers = **11**, was stale at 13; Multisig = 19) and `#[codec(compact)]` matches `#[pallet::compact]` (balances value compact, reversible amount fixed u128).
- Decodes the full `TxExtension` data after the call in declaration order: Era, `Compact<u32>` nonce, `Compact<u128>` tip, metadata-hash mode byte, spec/transaction version, genesis + block hash, `Option<[u8;32]>` metadata hash. Rejects trailing bytes and mode/Option inconsistency.
- Validates the genesis hash against a known-network whitelist (Planck `0x4901bf5c…`, Heisenberg `0xa5aa9e5c…`) and surfaces the network name; unknown genesis hard-fails.
- Multisig calls (create_multisig 0, propose 1, approve 2, execute 6): propose inner-call nesting capped at depth 2, inner call bytes must be fully consumed, total payload capped at 8 KB (nesting-bomb resistant).
- Display decimals fixed to 10^12; `ParsedPayload` Display surfaces network, era ("64 blocks" / "Immortal"), nonce, and tip.
- `bytes_to_ss58` takes `&[u8; 32]` — the `panic!` is gone.
- Removed unused `base58`/`blake2` deps.

### Dart display path (`quantus_sdk` + `cold-wallet-app`)
- `QuantusPayloadParser.parsePayload` now decodes the same full signed payload (call + extensions + trailing-byte/mode/genesis validation) and **throws `FormatException` with a clear reason** instead of returning null — no silent failures. Pallet index fixed 13 → 11; only `MultiAddress::Id` accepted; block-number delays rejected (apps only produce timestamp delays).
- `TransactionInfo.toString` decimals fixed 10^10 → `AppConstants.decimals` (12).
- Cold-wallet review screen shows parsed **Network, Tip, Nonce, Era** (never hardcoded zeros), formats the reversible window as a duration instead of mislabeled "blocks", and surfaces the exact rejection reason on the error screen.
- Multisig payloads (pallet 19) are still rejected by the Dart parser — the cold wallet has no multisig review UI, so they fail closed with a clear error (covered by a test).

### Tests (firmware suite shape)
- The two original "real world" vectors are kept verbatim as regression tests: both were captured on a retired devnet (genesis `0x826beefb…`). The transfer vector now fails with **unknown genesis** (proves the parser walks to the genesis hash); the reversible vector is rejected at call decode since its pallet index 13 was retired along with that devnet.
- Positive vectors reuse each original call portion (reversible re-indexed to pallet 11) plus a synthetic extension suffix: parameterized era/nonce/tip, spec 131, tx version 2, Planck genesis, arbitrary block hash, metadata None.
- Ported the firmware's negative suite: trailing bytes, bare call, metadata-mode mismatch, oversized payload, unknown pallet/call, huge vec length prefix, mortal era decode (0x8501 = period 64 phase 24), tip extraction, multisig depth limit + nesting bomb + inner-call trailing bytes.

## Test plan
- [x] `cargo test` in `rust-transaction-parser` — 18/18 pass
- [x] `flutter test` in `quantus_sdk` — 103 pass
- [x] `flutter test` in `cold-wallet-app` — 5 pass
- [x] `flutter test` in `mobile-app` — 202 pass
- [x] `flutter analyze --fatal-infos` clean on `quantus_sdk` and `cold-wallet-app`; `dart format` clean
- miner-app `widget_test.dart` fails to load on `main` too (no `main()` in file) — pre-existing, unrelated

### DIFF
--- a/cold-wallet-app/lib/screens/sign_transaction_screen.dart
+++ b/cold-wallet-app/lib/screens/sign_transaction_screen.dart
@@ -21,8 +21,8 @@ class SignTransactionScreen extends ConsumerStatefulWidget {
 }
 
 class _SignTransactionScreenState extends ConsumerState<SignTransactionScreen> {
-  TransactionInfo? _txInfo;
-  bool _parseFailed = false;
+  ParsedPayload? _parsed;
+  String? _parseError;
   String? _toCheckphrase;
   List<String>? _signatureUr;
   bool _signing = false;
@@ -31,13 +31,14 @@ class _SignTransactionScreenState extends ConsumerState<SignTransactionScreen> {
   @override
   void initState() {
     super.initState();
-    final info = QuantusPayloadParser.parsePayload(widget.payload);
-    if (info == null) {
-      _parseFailed = true;
-      return;
+    try {
+      final parsed = QuantusPayloadParser.parsePayload(widget.payload);
+      _parsed = parsed;
+      _loadCheckphrase(parsed.call.toAddress);
+    } catch (e) {
+      debugPrint('Rejected signing payload: $e');
+      _parseError = e is FormatException ? e.message : e.toString();
     }
-    _txInfo = info;
-    _loadCheckphrase(info.toAddress);
   }
 
   Future<void> _loadCheckphrase(String address) async {
@@ -86,12 +87,12 @@ class _SignTransactionScreenState extends ConsumerState<SignTransactionScreen> {
 
   @override
   Widget build(BuildContext context) {
-    if (_parseFailed) return _errorView(context);
+    if (_parseError != null) return _errorView(context, _parseError!);
     if (_signatureUr != null) return _signatureView(context, _signatureUr!);
-    return _reviewView(context, _txInfo!);
+    return _reviewView(context, _parsed!);
   }
 
-  Widget _errorView(BuildContext context) {
+  Widget _errorView(BuildContext context, String reason) {
     final colors = context.colors;
     final text = context.themeText;
     return ScaffoldBase(
@@ -108,6 +109,12 @@ class _SignTransactionScreenState extends ConsumerState<SignTransactionScreen> {
             style: text.smallParagraph?.copyWith(color: colors.textSecondary),
             textAlign: TextAlign.center,
           ),
+          const SizedBox(height: 12),
+          Text(
+            reason,
+            style: text.detail?.copyWith(color: colors.textMuted),
+            textAlign: TextAlign.center,
+          ),
         ],
       ),
       bottomContent: ScaffoldBaseBottomContent(
@@ -116,9 +123,11 @@ class _SignTransactionScreenState extends ConsumerState<SignTransactionScreen> {
     );
   }
 
-  Widget _reviewView(BuildContext context, TransactionInfo info) {
+  Widget _reviewView(BuildContext context, ParsedPayload parsed) {
     final colors = context.colors;
     final text = context.themeText;
+    final info = parsed.call;
+    final ext = parsed.extensions;
 
     return ScaffoldBase(
       appBar: const V2AppBar(title: 'Review & Sign'),
@@ -153,9 +162,17 @@ class _SignTransactionScreenState extends ConsumerState<SignTransactionScreen> {
             _detailRow(context, 'To', info.toAddress, monospace: true),
             if (_toCheckphrase != null && _toCheckphrase!.isNotEmpty)
               _detailRow(context, 'Checkphrase', _toCheckphrase!, valueColor: colors.checksum),
+            _detailRow(context, 'Network', parsed.network),
             _detailRow(context, 'Reversible', info.isReversible ? 'Yes' : 'No'),
             if (info.isReversible && info.reversibleTimeframe != null)
-              _detailRow(context, 'Reversible window', '${info.reversibleTimeframe} blocks'),
+              _detailRow(
+                context,
+                'Reversible window',
+                DatetimeFormattingService.formatDuration(Duration(milliseconds: info.reversibleTimeframe!)).formatted,
+              ),
+            _detailRow(context, 'Tip', '${_formatAmount(ext.tip)} ${AppConstants.tokenSymbol}'),
+            _detailRow(context, 'Nonce', '${ext.nonce}'),
+            _detailRow(context, 'Era', '${ext.era}'),
             if (_error != null) ...[
               const SizedBox(height: 16),
               Text(
--- a/quantus_sdk/lib/src/quantus_payload_parser.dart
+++ b/quantus_sdk/lib/src/quantus_payload_parser.dart
@@ -1,35 +1,96 @@
-/// A parser for Quantus blockchain transaction payloads.
+/// A parser for Quantus blockchain signing payloads.
 ///
-/// This parser extracts human-readable transaction information from SCALE-encoded
-/// payloads, specifically designed for hardware wallets that need to display
-/// transaction details to users before signing.
+/// Mirrors the Keystone firmware parser (rust/apps/quantus/src/parser.rs): the
+/// full signed payload — call plus every signed-extension field — is decoded
+/// with nothing left over, so what the signer displays is exactly what it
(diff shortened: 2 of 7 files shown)


cold-wallet-app/lib/screens/sign_transaction_screen.dart
```diff
@@ -153,9 +162,17 @@ class _SignTransactionScreenState extends ConsumerState<SignTransactionScreen> {
             _detailRow(context, 'To', info.toAddress, monospace: true),
```
dewabisma, line 162: I think we can wrap this in scrollable widget, so in small height phone it doesn't overflow.

quantus_sdk/lib/src/quantus_payload_parser.dart
```diff
@@ -40,154 +101,195 @@ class TransactionInfo {
 
   @override
   String toString() {
-    final amountStr = (amount / BigInt.from(10).pow(10)).toStringAsFixed(4);
+    final amountStr = (amount / BigInt.from(10).pow(AppConstants.decimals)).toStringAsFixed(4);
```
dewabisma, line 104: don't we have the formatting service for this exact thing? any reason not to reuse that instead?

dewabisma (changes requested):
One minor fix for UX and one confirmation needed.

n13 (commented):
## Review: parser criticals (full signed payload + network validation)

### Summary
This is a solid port of the Keystone firmware hardening. Both parsers now decode the **entire** signing payload (call + every TxExtension field) with a trailing-bytes check, fail closed on anything undeclared, and whitelist the genesis hash — closing the two audit criticals (partial parse / no network validation). I verified the decoding logic against the repo's generated chain metadata and live RPC, and ran all test suites.

### What I verified
- **Pallet/call indices** match the generated Planck bindings (`quantus_sdk/lib/generated/planck/`): Balances=2 (`transfer_allow_death`=0, `transfer_keep_alive`=3, compact value), ReversibleTransfers=11 (`schedule_transfer`=3, `schedule_transfer_with_delay`=4, **fixed** u128 amount), Multisig=19 (`create_multisig`=0, `propose`=1, `approve`=2, `execute`=6), `MultiAddress::Id`=0, `BlockNumberOrTimestamp` 0/1. The old pallet index 13 was indeed stale (13 is now TechCollective). Rust and Dart agree on every index and codec choice.
- **Extension layout** (era, compact nonce, compact tip, mode byte, spec u32, tx u32, genesis, block hash, `Option<metadata hash>`) matches exactly what the hot wallet's `QuantusSigningPayload.encodeRaw` emits via polkadart 0.7.3 for the generated registry's extension keys. Runtime nonce is u32 (`AccountInfo.nonce` is `U32Codec`), so `Compact<u32>` is right.
- **Era math** in `rust-transaction-parser/src/lib.rs` and `quantus_payload_parser.dart` is byte-for-byte identical to `sp_runtime::generic::Era::decode` including the `period >= 4 && phase < period` validity rule (spot-checked vectors: `8501` → Mortal(64, 24), `5501` → Mortal(64, 21)).
- **Genesis hashes confirmed against live chains**: `chain_getBlockHash(0)` returns `0x4901bf…5e72` on a1-planck.quantus.cat and `0xa5aa9e…3b4f` on a1/a2-heisenberg.quantus.cat — both match the whitelists.
- **Fail-closed behavior**: metadata mode/Option consistency enforced both sides; unknown pallet/call/address variants, bare calls, trailing bytes (incl. inside nested multisig calls), oversized payloads, and length-prefix bombs all rejected. The intentional asymmetry — Rust accepts Multisig (for firmware display), Dart rejects pallet 19 — is fail-closed for the cold wallet, which can't render multisig.
- **Cold wallet** (`sign_transaction_screen.dart`) only signs after a successful parse, signs the exact scanned bytes, and now shows network/tip/nonce/era; amount formatting uses exact BigInt math with `AppConstants.decimals` (12, matching the chain spec's `tokenDecimals`).

### Minor findings (non-blocking, display-only)
1. **Dart nonce range check can be bypassed by wrap-around** (`quantus_payload_parser.dart`, `_decodeExtensions`): `CompactCodec` decodes into a native 64-bit int, so a crafted compact value ≥ 2^63 wraps negative and slips past `nonce > 0xFFFFFFFF` (e.g. displays "Nonce: -1"). Such a payload is invalid on-chain (`Compact<u32>` rejects it), so nothing signable-and-executable is misdisplayed — but for strictness, decode with `CompactBigIntCodec` and range-check the BigInt, or also reject `nonce < 0`.
2. **`_decodeTimestampDelay` uses `BigInt.toInt()`**: a u64 timestamp ≥ 2^63 ms wraps negative before display. Absurd values only (≥ 292M years); consider rejecting delays above a sane bound instead of wrapping.
3. **No spec_version pinning**: neither parser constrains `spec_version`, so a future runtime upgrade that reshuffles pallet indices or TxExtension layout relies on decode-failure to fail closed (likely, given the trailing-byte check, but not guaranteed). Consider a minimum/known spec range as extra hardening — same trade-off the firmware makes, so fine to defer.
4. Nit: the Rust `Display` impl formats amounts via `f64`; exact values are preserved in the struct and the error is far below the 4 shown decimals for any plausible supply, but the Dart/BigInt approach is the better pattern if firmware ever displays large amounts.

### Test results (run locally on the PR head, a665f1a4)
- `cargo test` in `rust-transaction-parser`: **18/18 passed**.
- `flutter test test/quantus_payload_parser_test.dart` in `quantus_sdk`: **12/12 passed**.
- `flutter test` in `cold-wallet-app` (full suite): **5/5 passed**.
- `flutter test` full `quantus_sdk` suite: 3 test files fail in `setUpAll` (`generate_keys_test`, `migration_derivation_test`, `recovery_proxy_encoding_test`) because the `rust_lib_quantus_wallet` native library isn't built in this environment — **pre-existing**: identical failure on the merge base (552b6d67), unrelated to this PR.

**Verdict: Approve** (posted as comment — reviewer is PR author)

cold-wallet-app/lib/screens/sign_transaction_screen.dart
n13: It already is — the whole review view's `mainContent` is wrapped in `SingleChildScrollView` a few lines above this hunk (pre-existing from main, just outside the diff context). So the added Network/Tip/Nonce/Era rows scroll on short screens instead of overflowing.

quantus_sdk/lib/src/quantus_payload_parser.dart
n13: We do — `NumberFormattingService.formatBalance` — but it's deliberately not used here. This is `TransactionInfo.toString()`, which is only a debug/log string: nothing user-facing renders it (the cold-wallet review screen formats from the parsed fields directly). `formatBalance` is locale-aware (grouping/decimal separators come from the device locale) and smart-trims decimals, so its output varies by machine — wrong for a deterministic log string and for the fixed expectations in `quantus_payload_parser_test.dart`. The only change on this line was fixing the wrong scale (10^10 → `AppConstants.decimals`).

cold-wallet-app/lib/screens/sign_transaction_screen.dart
dewabisma: Awesome!

quantus_sdk/lib/src/quantus_payload_parser.dart
dewabisma: Okay, make sense!

dewabisma (approved):
LGTM!

PR #560, opened 2026-07-10
feat: multisig UX improvements

n13:
## Summary
- Fix fee calculation to show proposer's balance (not multisig balance) and display specific "Insufficient Balance for Fee" error
- Replace "TO" row with "PROPOSER" showing full address + checkphrase; remove all abbreviated addresses from review screen
- Client-side fallback decoding of proposal amount/recipient from `call_raw` when indexer fails
- Change "APPROVED" pill to "SIGNED" to avoid confusion about overall proposal status
- Add tilde prefix to estimated expiry time; enforce minimum 24h expiry
- Silent refresh for open proposals (no visible loading swirl)
- Add "DONE" button to dismiss proposal detail sheet after signing

## Test plan
- [x] Create a multisig proposal and verify proposer balance and fee error display correctly
- [x] Review screen shows full addresses and proposer row
- [x] Existing proposal with indexer decode failure still shows amount and recipient
- [x] Status pills show "SIGNED" not "APPROVED"
- [x] Open proposals refresh silently without loading indicator
- [x] After signing, "Done" button appears to dismiss the sheet


mobile-app/lib/v2/screens/send/input_amount_screen.dart
```diff
@@ -575,4 +585,29 @@ class _InputAmountScreenState extends ConsumerState<InputAmountScreen> {
...
+    AppLocalizations l10n,
+    String label,
+    ProviderListenable<AsyncValue<BigInt>> provider,
+    NumberFormattingService fmt,
+  ) {
+    final balanceAsync = ref.watch(provider);
+    return Row(
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        Text('$label ', style: text.smallParagraph?.copyWith(color: colors.textTertiary)),
```
dewabisma: We can extract the textTertiary as a variable instead of repeating in line.

dewabisma: Actually, since they also use the same style we can just extract the style as variable instead.

```
final textStyle = text.smallParagraph?.copyWith(color: colors.textTertiary);
```

dewabisma (approved):
LGTM just small nit.

mobile-app/lib/v2/screens/send/input_amount_screen.dart
n13: addressed

PR #561, opened 2026-07-10
feat: encrypted account send via ZK proofs

n13:
## Summary
- Add encrypted (wormhole) send flow with ZK proof generation, coin selection, and multi-batch submission
- New `EncryptedSendStrategy` integrating with the shared send UI via `SendStrategy` abstraction
- Wormhole coin selection algorithm with batch splitting for large UTXO sets
- Progress screen with step-by-step ZK proof generation feedback
- Account discovery service for gap-limit scanning of encrypted indices
- Refactor `wormhole_claim_service` → `wormhole_send_service` to support both claims and sends

## Test plan
- [ ] Send from encrypted account with sufficient balance completes successfully
- [ ] Insufficient encrypted funds shows appropriate blocker
- [ ] Batch below minimum exit threshold shows blocker
- [ ] Progress steps display correctly during proof generation
- [ ] Coin selection unit tests pass (`wormhole_coin_selection_test.dart`)
- [ ] Cancellation mid-proof returns to previous screen cleanly


n13 (commented):
Verdict: **Request changes**

I found two blocking correctness issues in the encrypted-send lifecycle:

1. **Preserve pending-spend state across refreshes** (`mobile-app/lib/shared/utils/polling_refresh_scope.dart:75`, `quantus_sdk/lib/src/services/encrypted_account_service.dart:92-100`). After pool acceptance, `onBatchSubmitted` persists nullifiers/change specifically to bridge the period before the indexer reports them. But pull-to-refresh and app-resume call `refreshActiveAccountBalance`, which calls `discardCachedState()` and deletes that same state file. If the indexer is still behind, the next `load()` exposes the just-spent inputs as spendable again and can also roll back/reuse `nextIndex`. Please clear only the chain-derived transfer/nullifier caches here and retain pending spends plus the monotonic address index until normal reconciliation/expiry.

2. **Keep cancellation scoped to the operation that was cancelled** (`quantus_sdk/lib/src/services/wormhole_send_service.dart:163-168`). `Future.any` returns `ClaimCancelled` while the proof flow keeps running, and its later `_checkCancelled()` calls read the mutable service-level `_cancelCompleter`. If the user closes the cancelled screen and starts another send before the orphaned FFI proof finishes, the second call replaces `_cancelCompleter`; the first flow now sees not-cancelled and can aggregate/submit after the UI already reported cancellation. Please capture/pass an operation-local cancellation token, or serialize operations, so an abandoned flow remains cancelled permanently.

Validation at head `c650e7b88c1c967bebf61833277bbc72153265d2`:
- SDK `flutter analyze --no-pub`: clean
- Mobile `flutter analyze`: two non-blocking `prefer_const_constructors` infos
- Focused wormhole coin-selection tests: 6/6 passed
- Rust library tests: 13/13 passed
- Full SDK suite: 97 tests passed; 3 setup failures because the local Flutter runner could not load `rust_lib_quantus_wallet.framework`
- Focused mobile tests could not build because the isolated worktree lacks the ignored `.env` asset

n13 (commented):
Verdict: **Request changes**

The pending-spend refresh issue from the previous review is fixed: refresh now preserves the persisted pending spends and monotonic `nextIndex` while clearing address-specific chain caches.

One blocking correctness issue remains:

1. **Cancellation still is not scoped to the cancelled operation** (`quantus_sdk/lib/src/services/wormhole_send_service.dart:106-113`, `quantus_sdk/lib/src/services/wormhole_send_service.dart:163-182`, `quantus_sdk/lib/src/services/wormhole_send_service.dart:193-195`). `cancel()` now awaits `_operationDone`, but `_operationDone` is completed by `_withCancellation`'s `finally`. Because `_withCancellation` awaits `Future.any`, that `finally` runs immediately when the cancellation guard wins; it does not wait for the orphaned `flow()` / FFI proof work to finish. The UI can therefore return from `cancel()`, allow a new send, and replace `_cancelCompleter`. The old flow's later `_checkCancelled()` and UTXO `isCancelled: () => _cancelled` calls then read the new operation's incomplete shared completer and can continue to aggregate or submit. `_opId` currently only guards cleanup and is never consulted by `_checkCancelled`, so it does not prevent this. Please give each flow an immutable operation-local token and pass/check that token throughout the flow (including the UTXO callback), or truly serialize operations by retaining/awaiting the underlying flow future before allowing another operation. Add a regression test that cancels during an unresolved proof, starts a second operation, completes the first proof, and verifies the first operation never reaches aggregation/submission.

Validation at head `46659e348be1767ee287876f7e30fbac6dbb3027`:
- SDK `flutter analyze --no-pub`: clean
- Focused wormhole coin-selection tests: 6/6 passed
- GitHub reports no checks for this branch

mobile-app/lib/v2/screens/send/encrypted_send_progress_screen.dart
```diff
@@ -0,0 +1,279 @@
...
+  int _currentStep = 0;
+  final Map<int, ClaimProgressItem> _stepProgress = {};
+
+  @override
+  void initState() {
+    super.initState();
+    _startSend();
+  }
+
+  Future<void> _startSend() async {
```
n13: **[bug]** `_startSend` captures `EncryptedAccountService` and runs proving/submit with no `dispose` cancellation. Logout clears wormhole caches + `encrypted_account_w*.json` (`SubstrateService.logout` / `EncryptedAccountService.clearAllPersistedState`) and only invalidates Riverpod providers (`logout_service.dart`), without calling `cancel()`. The in-flight `send` keeps the old service alive; `onBatchSubmitted` can recreate `encrypted_account_w{N}.json` after wipe, leaking pending nullifiers/change into the next session at the same wallet index. Secrets/mnemonic-derived material already held in memory also continue to be used post-logout.

**Suggestion:** On logout (and preferably in `EncryptedSendProgressScreen.dispose`), await `encryptedAccountService.cancel()` for every live wallet (or a process-wide send cancel registry) before clearing disk; make `onBatchSubmitted` / `_mutateState` no-op if a session generation token has changed; optionally refuse to persist after logout.

n13 (commented):
## Review by Grok 4.5

## Summary

This PR wires encrypted-account (wormhole) sends end-to-end: coin selection with volume fee / change, dual-exit ZK proofs via FRB, batch submit, Riverpod balance state, and a shared proving progress UI. The core spend math and FFI surface look coherent and are covered by solid coin-selection unit tests. Dominant risks are lifecycle races (logout / dispose while proving), UX promises that are not implemented (notifications, full private activity), and stale plans / nextIndex allocation that are not revalidated at submit time.

## Verdict

**REQUEST_CHANGES** (posted as COMMENT because GitHub disallows requesting changes on your own PR)

### Detail

REQUEST_CHANGES
Block on the logout/dispose race that can rewrite encrypted-account disk state after session wipe, and on the progress footer promising notifications that are never scheduled. Other items can follow as follow-ups but should not ship silently.

## Issue counts by severity

- bugs: 4
- suggestions: 4
- nits: 2

mobile-app/lib/v2/screens/send/encrypted_send_progress_screen.dart
```diff
@@ -0,0 +1,279 @@
...
+              child: LinearProgressIndicator(
+                value: _overallProgress,
+                backgroundColor: const Color(0xFF222222),
+                valueColor: AlwaysStoppedAnimation(colors.accentOrange),
+                minHeight: 4,
+              ),
+            ),
+            const SizedBox(height: 32),
+            Text(
+              _canceling ? l10n.commonCanceling : l10n.encryptedSendProgressFooter,
```
n13: **[bug]** Footer copy (`encryptedSendProgressFooter`) tells the user they can leave the app and will be notified when the private send completes. This flow never creates a `PendingTransactionEvent`, never starts pending-tx polling, and never posts a local/remote notification on success/failure. Proving also runs only while the isolate lives—OS kill mid-prove drops unsubmitted work with no recovery UX beyond a later balance refresh. `PopScope(canPop: !_running)` further prevents leaving the route while running (only backgrounding the app).

**Suggestion:** Either implement pending-state + notification on pool acceptance / finality (mirroring regular sends), or change the copy to match reality (must keep app open; no notification). Prefer recording submitted batch hashes and notifying when remaining work finishes or fails.

mobile-app/lib/v2/screens/send/encrypted_send_strategy.dart
```diff
@@ -0,0 +1,141 @@
...
+      encryptedBalanceProvider(account.walletIndex);
+
+  @override
+  bool extraBalancesLoading(WidgetRef ref) => false;
+
+  @override
+  BigInt feeChargedToBalance(SendFee? fee) => BigInt.zero;
+
+  @override
+  Future<SendFee> estimateFee(WidgetRef ref, {required String recipient, required BigInt amount}) async {
```
n13, line 61: **[bug]** Coin selection runs only in `estimateFee` and the resulting `WormholeSpendPlan` is reused all the way through review → auth → `EncryptedSendProgressScreen`. There is no re-load / re-select at submit time. If UTXOs become spent (another device, a prior partial batch, or a second concurrent flow), proofs fail late after long proving. Separately, `EncryptedAccountService.send` allocates `changeIndex` from `nextIndex` only at send start without reserving it; a parallel receive UI or second send can advertise/share the same address.

**Suggestion:** Re-run `load` + `selectWormholeInputs` immediately before proving (or at least validate planned nullifiers still unspent). Reserve/persist the change index when the plan is accepted (auth success), not only after first batch submit.

mobile-app/lib/v2/screens/activity/activity_screen.dart
```diff
@@ -160,7 +161,13 @@ class _ActivityScreenState extends ConsumerState<ActivityScreen> {
                         if (i > 0) const SizedBox(height: 32),
                         Text(group.label, style: text.receiveLabel?.copyWith(color: colors.textTertiary)),
                         ...group.transactions.mapIndexed((index, tx) {
-                          final itemData = TxItemData.from(tx, active.account.accountId, colors, l10n);
+                          final itemData = TxItemData.from(
+                            tx,
+                            active.account.accountId,
+                            colors,
+                            l10n,
+                            isPrivate: isEncryptedAccount(active.account),
```
n13, line 172: **[bug]** Private activity labels (`isPrivate: isEncryptedAccount(...)`) are applied to whatever GraphQL history is already loaded for `active.account.accountId`. Encrypted account identity is wormhole index 0 only (`AccountsService.createEncryptedAccount`), while receive/change rotate across HD indices. Deposits to the receive address and most private spends will not appear under index 0’s transfer history, so “Private Sent/Received” UI is largely cosmetic and activity remains empty or wrong for real encrypted usage.

**Suggestion:** Build encrypted activity from wormhole transfers / pending spends across discovered HD addresses (or a dedicated indexer view), not transparent `accountId` history alone. Until then, hide the activity list or show an explicit empty state explaining private activity is not listed.

quantus_sdk/lib/src/services/encrypted_account_service.dart
```diff
@@ -0,0 +1,327 @@
...
+
+  /// The address to show on the Receive screen: next unused index from the
+  /// last persisted state (cheap — no network). [load] keeps it current.
+  Future<WormholeKeyPair> receiveKeyPair() async => keyPairAt((await _readState()).nextIndex);
+
+  /// Drops on-disk transfer/nullifier caches for this wallet's known addresses
+  /// so the next [load] re-queries from chain. Preserves pending-spend records
+  /// and nextIndex — those are only pruned by [load]'s reconciliation or by
+  /// the 1-hour expiry, never by a refresh.
+  Future<void> discardCachedState() async {
```
n13, line 159: **[suggestion]** `discardCachedState` only clears caches for addresses already present in `_keyPairs`. After provider recreation (logout/login, process restart, first `encryptedAccountServiceProvider` read), `_keyPairs` is empty, so pull-to-refresh / `refreshActiveAccountBalance` becomes a no-op discard and relies entirely on existing on-disk transfer/nullifier caches. That is usually OK, but it cannot force a full cache drop when the in-memory key cache is cold.

**Suggestion:** Persist known address indices (or scan `nextIndex` + pending change addresses from `_FileState`) and clear caches for those even when `_keyPairs` is empty; or clear all wormhole caches for the wallet on force reload.

quantus_sdk/lib/src/services/encrypted_account_service.dart
```diff
@@ -0,0 +1,327 @@
...
+    final discoveredNext = usedIndices.isEmpty ? 0 : (usedIndices.reduce((a, b) => a > b ? a : b) + 1);
+    final state = await _mutateState((s) {
+      final kept = <PendingSpend>[];
+      for (final record in s.pendingSpends) {
+        final allSpent = record.nullifiers.every((n) => !unspentNullifiers.contains(n));
+        final changeArrived = record.changeAddress == null || usedAddresses.contains(record.changeAddress);
+        final age = DateTime.now().difference(DateTime.fromMillisecondsSinceEpoch(record.createdAtMs));
+        if (allSpent && changeArrived) {
+          _log('Pending spend confirmed on-chain, pruning (${record.nullifiers.length} nullifiers)');
+        } else if (age > _pendingSpendExpiry) {
```
n13, line 224: **[suggestion]** Pending spends older than 1 hour are dropped even if still unconfirmed. Under indexer lag, that can temporarily re-include spent UTXOs as spendable (failed double-spend attempts) or drop pending change from the displayed balance until discovery catches up. There is no user-visible error for the “ERROR: pending spend expired” log path.

**Suggestion:** Keep pending spends until nullifiers are observed spent (or until an explicit user refresh proves them spent); use expiry only as a soft warning. Surface a balance-reconciliation banner when expiry fires.

mobile-app/lib/v2/screens/send/encrypted_send_strategy.dart
```diff
@@ -0,0 +1,141 @@
...
+class EncryptedSendStrategy extends SendStrategy {
+  final Account account;
+
+  const EncryptedSendStrategy({required this.account});
+
+  @override
+  bool get showPrivateSendNotice => true;
+
+  @override
+  String? sourceAccountId(WidgetRef ref) => account.accountId;
```
n13, line 25: **[suggestion]** Self-send guard uses `account.accountId` (wormhole index 0 only). Users can still target their current receive address (`nextIndex`) or other HD wormhole addresses. That is wasteful (volume fee) and reduces privacy by linking addresses; depending on chain exit semantics it may also be surprising.

**Suggestion:** Treat all derived wormhole addresses for the wallet (at least `0..nextIndex` plus pending change addresses) as self-send destinations, or clearly allow “move to own address” as an explicit consolidation action.

quantus_sdk/test/services/wormhole_coin_selection_test.dart
```diff
@@ -0,0 +1,82 @@
...
+    transferCount: BigInt.one,
+  ),
+  owner: const WormholeAddressInfo(index: 0, address: 'addr', secretHex: '0x00'),
+  nullifierHex: '0xn$scaled',
+);
+
+BigInt quan(String v) => wormholePlanckFromScaled((double.parse(v) * 100).round());
+
+void main() {
+  group('selectWormholeInputs', () {
```
n13, line 23: **[suggestion]** Test coverage stops at pure coin selection. There are no tests for pending-spend reconciliation, `onBatchSubmitted` partial-batch persistence, dual-exit `WormholeLeafSpend` construction, logout/clear races, or encrypted fee blockers in the UI strategy.

**Suggestion:** Add unit tests for `EncryptedAccountService` state machine (pending prune/expiry, change index bump) and for `WormholeSendService.sendSpends` batch callback ordering; widget/strategy tests for quantization / insufficient / below-minimum blockers.

mobile-app/lib/v2/screens/send/encrypted_send_progress_screen.dart
```diff
@@ -0,0 +1,279 @@
...
+        mainContent: Column(
+          crossAxisAlignment: CrossAxisAlignment.stretch,
+          children: [
+            const SizedBox(height: 16),
+            _buildStatusHeader(colors, text, l10n),
+            const SizedBox(height: 24),
+            WormholeProgressSteps(
+              steps: [
+                (1, l10n.encryptedSendStepPreparing),
+                (2, l10n.encryptedSendStepGathering),
```
n13: **[nit]** Progress UI always lists steps 2–3 (“Gathering funds”, “Securing transaction”), but `sendSpends` only runs step 1 then jumps to 4–6. Those steps auto-complete when step 4 starts (`WormholeProgressSteps` via `_maxStartedStep`), which is slightly misleading during encrypted send (fine for redeem/claim).

**Suggestion:** Drive the step list from the operation type (claim vs send) or omit discovery steps for `sendSpends`.

quantus_sdk/lib/src/models/remote_config_model.dart
```diff
@@ -46,7 +46,7 @@ class RemoteConfigModel {
     enableHighSecurity: false,
     enableRemoteNotifications: true,
     enableSwap: false,
-    enableEncryptedAccount: false,
+    enableEncryptedAccount: true,
```
n13, line 49: **[nit]** Default remote config flips `enableEncryptedAccount` from `false` to `true`. That ships the feature to all clients without a remote kill-switch until config is fetched; combined with the lifecycle issues above this raises blast radius.

**Suggestion:** Keep default `false` until the logout/notification/activity gaps are closed, or ensure remote config is required before showing encrypted accounts.

quantus_sdk/test/services/wormhole_coin_selection_test.dart
n13: ok will add missing tests

quantus_sdk/lib/src/services/encrypted_account_service.dart
n13: not a bug

mobile-app/lib/v2/screens/send/encrypted_send_strategy.dart
n13: This is true. Treat all derived wormhole addresses for the wallet as self address.

quantus_sdk/lib/src/services/encrypted_account_service.dart
n13: force reload should clear all caches, that's the point

mobile-app/lib/v2/screens/activity/activity_screen.dart
n13: We can show total deposits and total spends, or else just show a private activity label. 

Solution: When we have established total balance, in that process we should have all data to show total received and total sent, so lets only show two items, total received, and total sent

Total received is across all wormhole addresses, total spent is the total of all nullifers, we already gather this information when calculating balance, so should be easy to put there, no additional calls needed.

mobile-app/lib/v2/screens/send/encrypted_send_strategy.dart
n13: No this is acceptable - this kind of race condition happens naturally in any blockchain on any tx, checking "again" doesn't fix it, and it's unlikely to happen in a real world scenario (1 user)

mobile-app/lib/v2/screens/send/encrypted_send_progress_screen.dart
n13: users are supposed to either wait for the send to go through or cancel the send

Action items:
- Make sure users are kept on the send page until complete or canceled
- if canceled, make sure we await the successful cancellation

mobile-app/lib/v2/screens/send/encrypted_send_progress_screen.dart
n13: change the footer text, they cannot leave the app

quantus_sdk/lib/src/services/encrypted_account_service.dart
n13: make sure force reload clears all caches

mobile-app/lib/v2/screens/send/encrypted_send_progress_screen.dart
n13: make sure our progress UX actually shows what we're doing!

n13 (commented):
## Review at head `2827618` — updated

Previous blocking items (pending-spend refresh, footer text, progress steps) are fixed. One remaining correctness issue found and patched in this review.

### Fix applied: cancellation state transition race

`_startSend()` was catching `ClaimCancelled` and setting `_running = false` immediately when `Future.any` picked up the cancel guard — but the underlying FFI flow was still running at that point. Since `PopScope(canPop: !_running)`, this allowed the user to navigate away before the operation had truly stopped.

Fix: `_startSend()` no longer transitions UI state on `ClaimCancelled`. Instead, `_cancel()` owns the transition — it awaits `cancel()` (which blocks until `_operationDone` resolves, i.e. the FFI work actually finishes), then sets `_running = false` / `_cancelled = true`. The screen stays in the "Canceling…" state with `canPop: false` until the operation has truly stopped.

This eliminates the theoretical race where a second operation could start while the first was still winding down.

### Non-blocking observations (no changes made)

1. **Self-send guard only checks index 0** — `EncryptedSendStrategy.sourceAccountId` returns `account.accountId` (wormhole index 0 only). The user can enter their current receive address (`nextIndex`) without triggering the self-send warning. Per PR discussion, all derived wormhole addresses for the wallet should be treated as self.

2. **`_readState()` outside `_stateLock`** — `send()` reads `nextIndex` outside the lock to allocate the change address. A concurrent `onBatchSubmitted` call (from the same send) could bump `nextIndex` between the read and usage. Low practical risk in single-isolate Dart but would be race-free if routed through `_mutateState`.

3. **Indonesian localization strings** — Several `encryptedSendStep*` entries in `app_localizations_id.dart` are still English. Not blocking.

### Validation

- SDK `flutter analyze --no-pub`: clean
- Focused wormhole coin-selection tests: 6/6 passed
- Encrypted account service tests: 3/3 passed

quantus_sdk/lib/src/services/encrypted_account_service.dart
```diff
@@ -0,0 +1,350 @@
...
+    }
+  }
+
+  Future<_FileState> _readState() async {
+    final file = await _stateFile();
+    if (!await file.exists()) return const _FileState(nextIndex: 0, pendingSpends: []);
+    return _FileState.fromJson(jsonDecode(await file.readAsString()) as Map<String, dynamic>);
+  }
+
+  Future<_FileState> _mutateState(_FileState Function(_FileState) fn) {
```
n13: **[bug]** Logout still races with in-flight encrypted-account mutations. `SubstrateService.logout()` calls `EncryptedAccountService.clearAllPersistedState()` (and wipes mnemonic/settings), but any concurrent `_mutateState` from `load()` / `send()` `onBatchSubmitted` keeps running on the old `EncryptedAccountService` instance. That path only does file I/O (no mnemonic check) and can recreate `encrypted_account_w{N}.json` after the wipe with the previous session’s `nextIndex` / `pendingSpends`. A subsequent wallet at the same `walletIndex` then inherits phantom pending change or a wrong next index. Riverpod invalidation after logout does not stop already-scheduled async work.

**Suggestion:** Add a generation/session token or `disposed` flag set by `clearAllPersistedState` / logout; every `_mutateState` must re-check it after acquiring `_stateLock` and before `writeAsString`. Prefer failing the write over resurrecting cleared state. Optionally `cancel()` in-flight sends and refuse new `load`/`send` after dispose.

quantus_sdk/lib/src/services/wormhole_send_service.dart
```diff
@@ -94,25 +118,73 @@ class WormholeClaimService {
...
+        _cancelCompleter = null;
+        _operationDone = null;
+      }
+      if (!done.isCompleted) done.complete();
+    });
+
+    try {
       final cancelGuard = cancelCompleter.future.then<ClaimResult>((_) => throw const ClaimCancelled());
-      return await Future.any([flow, cancelGuard]);
+      return await Future.any([flowFuture, cancelGuard]);
```
n13: **[bug]** Cancellation still fails open around submit. `_withCancellation` uses `Future.any([flowFuture, cancelGuard])`, so the caller receives `ClaimCancelled` as soon as `cancel()` completes the completer, while `flowFuture` continues until the next `_checkCancelled()`. There is no cancel check inside `Future.wait` proof generation or `_submitExtrinsic`. A cancel after the pre-submit `_checkCancelled()` (line 344) still submits the batch and may run `onBatchSubmitted`. `EncryptedSendProgressScreen` treats `ClaimCancelled` as non-success (only invalidates state; `_cancel` marks the UI cancelled), so the user can see “Send Cancelled” after funds actually left. Late cancel on the final batch is the worst case: full send succeeds, UI says cancelled.

**Suggestion:** Do not surface cancel to the UI via a racing `Future.any` alone. Either (a) only check cancel between batches and have `cancel()` await the true flow result, mapping “cancelled after N batches submitted” to a partial/success outcome, or (b) capture an op-local cancel flag and, when cancel wins the race, still await `flowFuture` and if `txHashes` is non-empty report partial/complete success instead of pure cancel. At minimum, re-check cancel immediately before `_submitExtrinsic` and document that in-flight submit cannot be aborted.

quantus_sdk/lib/src/services/encrypted_account_service.dart
```diff
@@ -0,0 +1,350 @@
...
+  /// batch lands) are persisted so balances stay exact even mid-flight or
+  /// after a partial failure.
+  Future<ClaimResult> send({
+    required WormholeSpendPlan plan,
+    required String recipientAddress,
+    required String circuitBinsDir,
+    required ClaimProgressCallback onProgress,
+    String? rpcUrl,
+  }) async {
+    final changeIndex = (await _readState()).nextIndex;
```
n13: **[suggestion]** Change-address allocation still does `final changeIndex = (await _readState()).nextIndex` outside `_stateLock`. Concurrent `_mutateState` writers (send batch persistence vs `load()` reconciliation after pull-to-refresh / provider invalidation) can interleave with this unlocked read. Writes are not atomic (direct `writeAsString`), so unlocked `_readState` can also observe a torn JSON file and fail the send. Index allocation is not reserved until a change-bearing batch’s `onBatchSubmitted` runs, so two overlapping sends could also pick the same change index.

**Suggestion:** Allocate and reserve `changeIndex` inside `_mutateState` at the start of `send` (bump or mark reserved under the lock). Route all state reads that inform writes through the lock; consider write-temp-then-rename for the state file.

mobile-app/lib/v2/screens/send/encrypted_send_strategy.dart
```diff
@@ -0,0 +1,140 @@
...
```
n13, line 25: **[suggestion]** Self-send protection still only compares the recipient to `account.accountId`, which for encrypted accounts is the wormhole address at HD index 0 (`AccountsService.createEncryptedAccount`). Deposits/change use later indices shown on Receive (`receiveKeyPair()` → `nextIndex`). Pasting your current receive address or an older change address is allowed, burning the privacy fee for a self-transfer with no warning.

**Suggestion:** For encrypted strategy, treat any derived wormhole address in `0..nextIndex` (or known `_keyPairs` / discovered indices) as self-send, or resolve the recipient against the wallet’s wormhole address set before enabling Continue.

mobile-app/lib/v2/screens/send/encrypted_send_strategy.dart
```diff
@@ -0,0 +1,140 @@
...
+  Future<SendOutcome> submit(
+    WidgetRef ref, {
+    required String recipientAddress,
+    required String recipientChecksum,
+    required BigInt amount,
+    required SendFee fee,
+    required bool isPayMode,
+  }) async {
+    final l10n = ref.read(l10nProvider);
+    final plan = (fee as EncryptedFee).plan;
```
n13, line 127: **[suggestion]** Submit uses the `EncryptedFee.plan` frozen at fee-estimate time on the amount screen with no revalidation. UTXO spendability can change before confirm (background invalidation/load, another partial send’s pending spends, long dwell on review). The progress screen then proves a stale plan; failures surface only after expensive proving, and there is no refresh of coin selection immediately before auth/prove.

**Suggestion:** On confirm (or at the start of `EncryptedSendProgressScreen`), re-load encrypted state and rebuild/verify the plan (same amount + still-unspent nullifiers). If inputs are gone, fail fast with a clear “balances changed, re-enter amount” error before circuit work.

n13 (commented):
Verdict: **Request changes**

GitHub will record this as a comment because the authenticated reviewer account owns the PR.

At head `5f61718`, the earlier pending-spend refresh, progress-footer, derived-address self-send, encrypted activity summary, and Indonesian localization findings are fixed. The coin-selection and dual-exit wiring look coherent, mobile analysis is clean, the focused SDK suite passes 22/22, and the GitHub Analyze check is green.

I still see two blocking lifecycle issues:

1. **Logout does not quiesce encrypted-account work before wiping the session** (`mobile-app/lib/services/logout_service.dart:39-58`, `quantus_sdk/lib/src/services/substrate_service.dart:391-398`, `quantus_sdk/lib/src/services/encrypted_account_service.dart:137-191, 238-258, 305-313`). Invalidating Riverpod providers does not cancel an already-running `load()` or `send()`. Both retain the old mnemonic-derived material; a delayed `load()` can write `_FileState` after logout, and a delayed submitted batch can call `onBatchSubmitted` and recreate `encrypted_account_w{N}.json` after `clearAllPersistedState()`. That lets old-wallet state leak into the next session at the same wallet index, and a send already past authentication can continue after logout. Please cancel and await every live encrypted operation before clearing settings/caches, and gate every post-await persistence callback with a session-generation token. Add a delayed-load/delayed-send versus logout regression test.

2. **Cancellation is still service-global rather than operation-local** (`quantus_sdk/lib/src/services/wormhole_send_service.dart:97-112, 169-205, 232-239`). `_withCancellation` creates a local `cancelCompleter`, but `_checkCancelled()` and the UTXO callback read `_cancelled`, which dereferences the current service-level `_cancelCompleter`. `_opId` only protects cleanup. If a new operation starts on the same service before the old FFI flow reaches a checkpoint, it replaces the completer and the cancelled flow can resume aggregation/submission. Awaiting `cancel()` in the encrypted progress screen mitigates one caller, but it does not make the shared service safe and the miner/redeem callers do not await cancellation. Thread an immutable per-operation context/token through the flow and have `cancel()` await the flow it actually cancelled. Add the unresolved-proof / cancel / start-second-op regression test requested in the previous review.

Before merge, I would also address these repository-policy and user-safety items:

- `EncryptedSendProgressScreen` owns the send state machine and launches it from `initState` (`mobile-app/lib/v2/screens/send/encrypted_send_progress_screen.dart:44-98`). Quantus guidance requires state in providers and widgets to be renderers; this lifecycle ownership is also what makes disposal/logout hard to control. Move the operation/controller state into an auto-disposed provider and cancel/await it on disposal/session teardown.
- The copy `Activity not visible on chain` (`mobile-app/lib/l10n/app_en.arb:2810`) overstates the privacy guarantee. Wormhole hides source linkage, but exits still create observable recipient/amount state changes. Describe the actual guarantee instead of claiming the activity is absent from chain.
- The new progress screen hardcodes colors and text styles and uses `withValues(alpha:)` (`mobile-app/lib/v2/screens/send/encrypted_send_progress_screen.dart:194-216, 224-239, 285-299`). The design-system rules require `context.colors.*`, `context.themeText.*`, and `.useOpacity()`.

Validation at `5f617181a34b2da46933c4c461eb82d7e956c827`:

- GitHub Analyze: pass
- `mobile-app`: `flutter analyze` — no issues
- `quantus_sdk`: focused coin-selection + encrypted-account-service tests — 22/22 passed

miner-app/lib/features/withdrawal/claim_rewards_dialog.dart
```diff
@@ -29,7 +29,7 @@ enum _Screen { input, confirm, progress }
 
 class _ClaimRewardsDialogState extends State<_ClaimRewardsDialog> {
   final _addressController = TextEditingController();
-  final _claimService = WormholeClaimService();
+  final _claimService = WormholeSendService();
```
dewabisma: This naming is seriously conflicting, we need to change to better name.

illuzen: i think the two services are merged

n13: We renamed it to sendservice, there's only 1

illuzen (approved):

mobile-app/lib/services/logout_service.dart
```diff
@@ -50,6 +50,12 @@ class LogoutService {
     _ref.read(activeAccountProvider.notifier).reset();
     _ref.invalidate(recoveryPhraseViewedProvider);
     _ref.invalidate(walletOriginProvider);
+    // Drop encrypted-account Riverpod state so the next session doesn't show
+    // the previous wallet's balance while disk caches are already wiped above.
+    _ref.invalidate(encryptedAccountServiceProvider);
+    _ref.invalidate(encryptedStateProvider);
+    _ref.invalidate(encryptedBalanceProvider);
+    _ref.invalidate(encryptedSpendableProvider);
```
dewabisma, line 58: Why total spent provider not being reset as well?

miner-app/lib/features/withdrawal/claim_rewards_dialog.dart
dewabisma: Okay, just weird to say it send service only to name the variable claimService

mobile-app/lib/v2/components/wormhole_progress_steps.dart
```diff
@@ -0,0 +1,132 @@
...
+      isCompleted = reachedTotal;
+      isActive = progress != null && !reachedTotal && !cancelled && !hasError;
+    } else {
+      isCompleted = _maxStartedStep > step;
+      isActive = !isCompleted && !cancelled && !hasError && currentStep == step;
+    }
+
+    final Widget icon;
+    if (isCompleted) {
+      icon = SizedBox(
```
dewabisma: This is repeating code, clearly violation of DRY.

dewabisma: Many repeated

mobile-app/lib/v2/components/wormhole_progress_steps.dart
```diff
@@ -0,0 +1,132 @@
...
+      );
+    } else {
+      icon = SizedBox(
+        width: 22,
+        height: 22,
+        child: Center(
+          child: Container(
+            width: 5,
+            height: 5,
+            decoration: BoxDecoration(color: const Color(0xFF222222), borderRadius: BorderRadius.circular(2.5)),
```
dewabisma: If this is new color and will be reused, let's add it to the color theme

mobile-app/lib/v2/components/wormhole_progress_steps.dart
```diff
@@ -0,0 +1,132 @@
...
+          ),
+        ),
+      );
+    }
+
+    const pendingColor = Color(0xFF666666);
+    final titleColor = isCompleted
```
dewabisma: As usual, I don't like reading chain ternary.

mobile-app/lib/v2/screens/accounts/account_ready_screen.dart
```diff
@@ -90,44 +89,62 @@ class AccountReadyScreen extends ConsumerWidget {
...
+                            children: [
+                              ref
+                                  .watch(checksumNameProvider(accountId))
+                                  .when(
+                                    data: (checksum) => Text(
+                                      checksum,
+                                      textAlign: TextAlign.center,
+                                      style: text.smallParagraph?.copyWith(
+                                        color: colors.checksum,
+                                        fontWeight: FontWeight.w400,
```
dewabisma: smallParagraph is already w400, so it's unnecessary to override it

mobile-app/lib/v2/screens/accounts/account_ready_screen.dart
```diff
@@ -147,3 +164,78 @@ class AccountReadyScreen extends ConsumerWidget {
...
+        children: [
+          leading,
+          const SizedBox(width: 12),
+          Expanded(
+            child: Column(
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: [
+                Text(title, style: text.paragraph?.copyWith(fontSize: 18, height: 1.0)),
+                const SizedBox(height: 4),
+                Text(description, style: text.detail?.copyWith(color: colors.textMuted, fontSize: 12, height: 1.0)),
```
dewabisma: I don't know why AI like to override the text style when the font size defined in textTheme already the same value.

Image

mobile-app/lib/v2/screens/import/import_wallet_screen.dart
```diff
@@ -147,12 +147,30 @@ class _ImportWalletScreenV2State extends ConsumerState<ImportWalletScreenV2> {
       }
       ref.invalidate(accountsProvider);
       ref.invalidate(activeAccountProvider);
+      await _discoverEncryptedAccount();
```
dewabisma: is waiting here preferred? We seems to not want to await on initial flow either create or import, why do this now? 

I think we should just do in background and show that private account is syncing or preparing, instead of waiting in import phase.

dewabisma (changes requested):
DRY need to be addressed before merge.

dewabisma (approved):
Please address the one that is reasonable to fix. I think checking on the provider invalidation and DRY violation is reasonable.

mobile-app/lib/v2/screens/import/import_wallet_screen.dart
n13: Hmm yeah we don't want to block wallet creation!

mobile-app/lib/v2/components/wormhole_progress_steps.dart
n13: True

mobile-app/lib/v2/components/wormhole_progress_steps.dart
n13: Ok will check this

mobile-app/lib/v2/components/wormhole_progress_steps.dart
n13: I like it ;)

What do you suggest that's better? Switch?

mobile-app/lib/v2/screens/accounts/account_ready_screen.dart
n13: True - should not override text style defaults with the same value

Maybe for the font size at times, but we should avoid as much as possible and this one seems frivolous

mobile-app/lib/v2/components/wormhole_progress_steps.dart
dewabisma: Not really switch. We can have a method instead for this. So we can have normal if else block instead of chaining ternary.

Nothing serious just my own preference on code style.

dewabisma (approved):
LGTM!

PR #563, opened 2026-07-11
Approve multisig proposals with any local signer, Add keystone support

n13:
## Summary
- Fixes [#562](https://github.com/Quantus-Network/quantus-apps/issues/562): when multiple local accounts are signers of the same multisig, the app only tracked `myMemberAccountId`, so after that account signed there was no way for another local co-signer to approve.
- Approval eligibility now considers **all local accounts that are multisig members**.
- If more than one eligible local signer remains, tapping **Approve** opens a picker listing only accounts that have not already signed (or have a pending approval).
- **Keystone hardware signers** no longer hit `Mnemonic not found for signing` — multisig approve/execute/cancel detect Keystone accounts and use the shared QR sign → scan signature → external submit path (same pattern as transfers).

## Changes
- Multi-local-signer helpers + unit tests
- Signer picker sheet + EN/ID l10n
- Proposal detail / approve confirm / activity row updates
- Generic `KeystoneExtrinsicSession` + sign/scan screens
- Hardware submit methods for multisig approve/execute/cancel

## Test plan
- [x] Add 2 accounts on one device and create a 2-of-2 (or 2-of-N) multisig with both as signers
- [x] Create a proposal; first account auto-approves as proposer (or approve as account 1)
- [x] Open the proposal while account 1 has already signed — **Approve** remains available
- [x] With only one remaining unsigned local software signer, Approve goes straight to confirm and signs as that account
- [x] With two unsigned local signers, Approve shows the picker and excludes accounts that already signed
- [x] With a **Keystone** co-signer selected: confirm → Keystone QR screen → scan signature QR → approval submitted / pending state shown
- [x] Execute / cancel with a Keystone member also uses the QR flow
- [x] After all local signers have approved, button shows **Already Signed** / Done as before
- [x] Single-local-signer software multisigs still behave as before (no picker)


mobile-app/lib/services/transaction_submission_service.dart
```diff
@@ -242,6 +242,44 @@ class TransactionSubmissionService {
     }
   }
 
+  /// Approves a proposal using a signature produced off-device (Keystone).
```
dewabisma, line 222: I think this new external signature function can be refactored because it share repeating code, with only the data constructor is different.

mobile-app/lib/v2/screens/multisig/multisig_action_confirm_sheet.dart
```diff
@@ -100,6 +142,10 @@ class _MultisigActionConfirmSheetState extends ConsumerState<MultisigActionConfi
     return signer;
   }
 
+  bool _isHardwareSigner(Account signer) {
```
dewabisma, line 146: I think we already have the extension for checking if account is hardware wallet or high security or other type of account, so we don't need to create new function everytime

mobile-app/lib/v2/screens/multisig/multisig_action_confirm_sheet.dart
```diff
@@ -153,6 +207,48 @@ class _MultisigActionConfirmSheetState extends ConsumerState<MultisigActionConfi
...
+    );
+
+    setState(() => _submitting = false);
+
+    final ok = await Navigator.of(
+      context,
+    ).push<bool>(MaterialPageRoute(builder: (_) => KeystoneSignExtrinsicScreen(session: session)));
+
+    if (!mounted) return;
+    if (ok == true) {
```
dewabisma: Is this already showing toast or some feedback if not ok?

mobile-app/lib/v2/screens/multisig/multisig_activity_section.dart
```diff
@@ -38,14 +40,18 @@ class MultisigActivitySection extends ConsumerWidget {
     final openProposalsAsync = ref.watch(multisigOpenProposalsProvider(msig));
     final pastProposalsAsync = ref.watch(multisigPastProposalsProvider(msig));
     final pending = pendingProposalsForMultisig(ref.watch(pendingMultisigProposalsProvider), msig.accountId);
+    final accounts = ref.watch(accountsProvider).value ?? const <Account>[];
+    final localSignerIds = localSignersForMultisig(msig, accounts).map((a) => a.accountId).toList();
```
dewabisma, line 44: We seems to passing this around, maybe we can have a proper provider for local signers multisig? The current approach seems clunky.

mobile-app/lib/v2/screens/multisig/multisig_signer_picker_sheet.dart
```diff
@@ -0,0 +1,90 @@
...
+          children: [
+            AccountBadge.account(account: account, size: 40),
+            const SizedBox(width: 12),
+            Expanded(
+              child: Column(
+                crossAxisAlignment: CrossAxisAlignment.start,
+                children: [
+                  Text(
+                    account.name,
+                    style: text.smallParagraph?.copyWith(color: colors.textPrimary, fontWeight: FontWeight.w600),
```
dewabisma, line 69: property color is unnecessary since we use textPrimary by default.

mobile-app/lib/v2/screens/send/keystone_scan_extrinsic_signature_screen.dart
```diff
@@ -0,0 +1,245 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final l10n = ref.watch(l10nProvider);
+    final colors = context.colors;
+    final text = context.themeText;
+    final size = MediaQuery.of(context).size;
+    final frame = (size.width - 96).clamp(220.0, 300.0);
+    final progress = _expectedParts == null ? null : (_seenSeq.length / _expectedParts!).clamp(0.0, 1.0);
+
+    return Scaffold(
```
dewabisma: I remember we already have this keystone scan like screen, can't we refactor it to be reusable instead of creating totally new screen?

mobile-app/lib/v2/screens/send/keystone_sign_extrinsic_screen.dart
```diff
@@ -0,0 +1,186 @@
...
+    );
+  }
+
+  Widget _details(AppColorsV2 colors, AppTextTheme text, KeystoneExtrinsicSession session) {
+    return Column(
+      children: [
+        if (session.primaryDetail != null)
+          Text(
+            session.primaryDetail!,
+            style: text.smallTitle?.copyWith(color: colors.textPrimary),
```
dewabisma: unnecessary color override

mobile-app/lib/v2/screens/send/keystone_sign_extrinsic_screen.dart
```diff
@@ -0,0 +1,186 @@
...
+            textAlign: TextAlign.center,
+          ),
+        if (session.secondaryDetail != null) ...[
+          const SizedBox(height: 8),
+          Text(
+            session.secondaryDetail!,
+            style: text.detail?.copyWith(
```
dewabisma: I think the font size override is unecessary, because detail already font size 12 AFAIK, and we have also other text style that match the fontSize and maybe family as well instead of override here.

dewabisma (changes requested):
So, does it mean when we have multiple eligible signers, we have to sign twice in the app now?

Some changes needed.

mobile-app/lib/v2/screens/send/keystone_scan_extrinsic_signature_screen.dart
n13: This is pretty bad. 

Fixed.

mobile-app/lib/v2/screens/multisig/multisig_action_confirm_sheet.dart
n13: There is no toast in this parent when ok is null/false. In this flow, true means the signed extrinsic was submitted successfully; null means the user backed out, so no feedback is needed. Preparation and scan/submission failures stay on the Keystone screens: they show an inline localized error, and the scanner resets/restarts so the user can retry.

mobile-app/lib/services/transaction_submission_service.dart
n13: I blame the AI for this DRY violation. Grrr.

mobile-app/lib/v2/screens/multisig/multisig_action_confirm_sheet.dart
n13: this is a small utility method for debugging, it's ok

mobile-app/lib/v2/screens/multisig/multisig_activity_section.dart
n13: Hmm maybe but it's a bit of a niche case of a niche case

mobile-app/lib/v2/screens/send/keystone_sign_screen.dart
```diff
@@ -186,6 +132,39 @@ class _KeystoneSignScreenState extends ConsumerState<KeystoneSignScreen> {
     );
   }
 
+  Widget _details(AppColorsV2 colors, AppTextTheme text, KeystoneSigningSession session) {
+    return Column(
+      children: [
+        if (session.primaryDetail != null)
+          Text(
+            session.primaryDetail!,
+            style: text.smallTitle?.copyWith(color: colors.textPrimary),
```
dewabisma, line 141: Unnecessary color override.

mobile-app/lib/v2/screens/send/keystone_sign_screen.dart
```diff
@@ -186,6 +132,39 @@ class _KeystoneSignScreenState extends ConsumerState<KeystoneSignScreen> {
...
+            textAlign: TextAlign.center,
+          ),
+        if (session.secondaryDetail != null) ...[
+          const SizedBox(height: 8),
+          Text(
+            session.secondaryDetail!,
+            style: text.detail?.copyWith(
+              color: colors.textMuted,
+              fontFamily: AppTextTheme.fontFamilySecondary,
+              fontSize: 12,
```
dewabisma, line 151: Unnecessary fontSize override

dewabisma (approved):
Looking great! Just few nits.

PR #564, opened 2026-07-14
Implement account menu design

n13:
Implement account menu designs
Simulator Screenshot - iPhone 17 - 2026-07-14 at 12 51 50
Simulator Screenshot - iPhone 17 - 2026-07-14 at 12 51 54


mobile-app/lib/v2/screens/settings/add_account_menu_screen.dart
```diff
@@ -0,0 +1,287 @@
...
+            text: text,
+          ),
+          const SizedBox(height: 14),
+          if (enableKeystone) ...[
+            _AccountOptionCard(
+              icon: Icons.qr_code_scanner,
+              title: l10n.addAccountMenuImportKeystoneTitle,
+              subtitle: l10n.addAccountMenuImportKeystoneSubtitle,
+              onTap: _onImportKeystone,
+              colors: colors,
```
dewabisma, line 91: Since this widget is a proper widget instead of a var or just method. I think we can get the colors and text inside the widget definition instead of passing as parameters. This way we don't repeat unnecessarily.

mobile-app/lib/v2/screens/settings/add_account_menu_screen.dart
```diff
@@ -0,0 +1,287 @@
...
+              text: text,
+              children: [
+                GestureDetector(
+                  onTap: _onDiscoverMultisig,
+                  behavior: HitTestBehavior.opaque,
+                  child: _AccountOptionRowContent(
+                    icon: Icons.radar_outlined,
+                    title: l10n.addAccountMenuDiscoverMultisigTitle,
+                    subtitle: l10n.addAccountMenuDiscoverMultisigSubtitle,
+                    colors: colors,
```
dewabisma, line 123: Same issue here with repeating colors and text theme.

mobile-app/lib/v2/screens/settings/add_account_menu_screen.dart
```diff
@@ -0,0 +1,287 @@
...
+      padding: const EdgeInsets.all(16),
+      child: Column(
+        children: [
+          GestureDetector(
+            onTap: onToggle,
+            behavior: HitTestBehavior.opaque,
+            child: Row(
+              mainAxisAlignment: MainAxisAlignment.spaceBetween,
+              children: [
+                Text(title, style: text.paragraph?.copyWith(fontSize: 18, color: colors.textPrimary)),
```
dewabisma, line 257: AI really like to do unnecessary override like `colors.textPrimary`.

We don't need to, text primary is the default color.

dewabisma (approved):
LGTM! Just few nits.

mobile-app/lib/v2/screens/settings/add_account_menu_screen.dart
n13: theyre not uneccessary

PR #566, opened 2026-07-21
always render swap

n13:
just disabled on the encrypted account

this is a better UX - we don't want to have buttons appear and disappear


dewabisma (approved):
👍

PR #569, opened 2026-07-22
fix: lock lifecycle fails closed on process death (M1)

n13:
## Summary

Fixes finding **M1** from the security audit: the app-lock lifecycle could be bypassed and failed open. Stacked on top of #567 (H1).

Three defects, three fixes:

1. **Grace window survived process death.** The `last_paused_time` timestamp lived in SharedPreferences, so background → swipe-kill → relaunch within 10 s came back fully unlocked. The timestamp is now **in-memory only** (`LocalAuthService._lastPausedTime`); any process death (kill, crash, reboot) resets it.
2. **Null timestamp failed open.** `shouldRequireAuthentication()` returned `false` when no timestamp existed, and the timestamp was cleaned on every successful auth — so crash/force-stop/reboot after an unlock meant relaunching with no auth, no time limit. A null pause record with an existing wallet now means **locked** (fail closed).
3. **Unlocked-by-default initial state.** `LocalAuthState` defaulted to `isAuthenticated: true`, rendering content before the async check completed. It now defaults to locked; `AuthWrapper` shows the lock screen until auth succeeds.

Additionally, a bypass found while fixing the above: `checkAuthentication()`'s "within grace window" branch set `isAuthenticated: true` **even for sessions that had never been unlocked** — backgrounding from the lock screen and resuming within the grace window would have unlocked the app with no prompt (with defects 1–2 fixed, this would have become the easiest way in). The grace window now only applies to sessions that were actually authenticated in this process; a locked session always goes through `authenticate()`.

The SDK's now-dead `SettingsService` timestamp methods (`getLastPausedTime`/`setLastPausedTime`/`cleanLastPausedTime` and the `last_paused_time` SharedPreferences key) are removed in this PR as well — the mobile app was their only consumer.

## Behavior notes

- The 10-second background grace still works as before **within a process** for unlocked sessions (background → quick resume doesn't re-prompt).
- Cold start with a wallet now always prompts (previously it silently unlocked whenever the timestamp was missing) — on devices with lock-screen security enrolled.
- Devices with **no lock-screen security** (no PIN/pattern/biometric) are not gated: no prompt, no lock screen, gated actions pass. There is nothing to enforce on such a device, and failing closed would just dead-end the app on a lock screen whose only affordance retries an auth that can never succeed.

## Testing

- `test/unit/local_auth_service_test.dart` extended to 15 tests: fail-closed on null pause record, grace-window boundaries (via an injectable `authTimeout`), pause-record cleared only on successful auth, controller starts locked, locked-session grace bypass regression, unlocked-session grace behavior.
- `flutter analyze lib` clean; full `test/unit` suite passes (220 tests).
- Not executed on hardware: lifecycle edge cases (force-stop, reboot, biometric-prompt oscillation guard at `app_lifecycle_manager.dart:117`) are worth a manual smoke test on both platforms.

---

#### M1. Lock lifecycle: 10-second grace survives process death; null timestamp fails open; unlocked-by-default initial state
`mobile-app/lib/services/local_auth_service.dart:14,55,66-76`; `lib/providers/local_auth_provider.dart:12,64`; `lib/app_lifecycle_manager.dart:117-122`

The lock decision rests entirely on a `last_paused_time` SharedPreferences timestamp. Three compounding defects:

1. `shouldRequireAuthentication()` returns `false` when the timestamp is null (`local_auth_service.dart:69-70`), and `cleanLastPausedTime()` runs on *every successful auth* (line 55). Crash, foreground force-stop, or reboot with the app open → relaunch is unlocked with no time limit.
2. **Trivial bypass (both platforms, no tools):** background → swipe-kill → relaunch within the 10 s `_authTimeout` → fully unlocked. The grace window should not survive process death.
3. `LocalAuthState` defaults to `isAuthenticated: true` (`local_auth_provider.dart:12`), so content renders before the async check completes, and the lock is only a widget overlay (`lib/app.dart:55`); initialization and polling run regardless.

**Mitigating factor (verified):** per-action gates call `authenticate()` directly, so on biometric-enrolled devices funds and seed remain behind a prompt even when the app opens unlocked; the bypass primarily exposes balances/history/addresses. On PIN-only devices it chains with H1 into full access.
**Fix (this PR):** persist lock state in memory rather than inferring it from a missing timestamp; treat "wallet exists + no recorded pause" as locked; initialize `isAuthenticated = false`; require re-auth after any process death regardless of elapsed time.

### DIFF
--- a/mobile-app/lib/services/local_auth_service.dart
+++ b/mobile-app/lib/services/local_auth_service.dart
@@ -9,15 +9,26 @@ class LocalAuthService {
 
   final LocalAuthentication _localAuth;
   final SettingsService _settingsService;
+  final Duration _authTimeout;
 
-  LocalAuthService._internal() : _localAuth = LocalAuthentication(), _settingsService = SettingsService();
+  /// In-memory only: the background grace window must never survive
+  /// process death. A null value with an existing wallet therefore means
+  /// "this process has not recorded a backgrounding" — i.e. locked.
+  DateTime? _lastPausedTime;
+
+  LocalAuthService._internal()
+    : _localAuth = LocalAuthentication(),
+      _settingsService = SettingsService(),
+      _authTimeout = const Duration(seconds: 10);
 
   @visibleForTesting
-  LocalAuthService.withDependencies({required LocalAuthentication localAuth, required SettingsService settingsService})
-    : _localAuth = localAuth,
-      _settingsService = settingsService;
-
-  static const _authTimeout = Duration(seconds: 10);
+  LocalAuthService.withDependencies({
+    required LocalAuthentication localAuth,
+    required SettingsService settingsService,
+    Duration authTimeout = const Duration(seconds: 10),
+  }) : _localAuth = localAuth,
+       _settingsService = settingsService,
+       _authTimeout = authTimeout;
 
   /// Whether the device has any lock-screen security enrolled: biometrics
   /// or a device credential (PIN/pattern/password).
@@ -87,8 +98,10 @@ class LocalAuthService {
   Future<bool> shouldRequireAuthentication() async {
     try {
       if (!await _settingsService.getHasWallet()) return false;
-      final lastPausedTime = _settingsService.getLastPausedTime();
-      if (lastPausedTime == null) return false;
+      // Fail closed: a wallet with no in-memory pause record means the
+      // process started (or was killed) since the last unlock.
+      final lastPausedTime = _lastPausedTime;
+      if (lastPausedTime == null) return true;
       return DateTime.now().difference(lastPausedTime) > _authTimeout;
     } catch (e) {
       debugPrint('Error checking if authentication is required: $e');
@@ -98,12 +111,12 @@ class LocalAuthService {
 
   Future<bool> updateLastPausedTime() async {
     if (!await _settingsService.getHasWallet()) return false;
-    _settingsService.setLastPausedTime(DateTime.now());
+    _lastPausedTime = DateTime.now();
     return true;
   }
 
   void cleanLastPausedTime() {
-    _settingsService.cleanLastPausedTime();
+    _lastPausedTime = null;
   }
 
   Future<void> stopAuthentication() async {
--- a/mobile-app/lib/services/local_auth_service.dart
+++ b/mobile-app/lib/services/local_auth_service.dart
@@ -6,12 +6,29 @@ import 'package:quantus_sdk/quantus_sdk.dart';
 class LocalAuthService {
   static final LocalAuthService _instance = LocalAuthService._internal();
   factory LocalAuthService() => _instance;
-  LocalAuthService._internal();
 
-  final LocalAuthentication _localAuth = LocalAuthentication();
-  final SettingsService _settingsService = SettingsService();
+  final LocalAuthentication _localAuth;
+  final SettingsService _settingsService;
+
+  LocalAuthService._internal() : _localAuth = LocalAuthentication(), _settingsService = SettingsService();
+
+  @visibleForTesting
+  LocalAuthService.withDependencies({required LocalAuthentication localAuth, required SettingsService settingsService})
+    : _localAuth = localAuth,
+      _settingsService = settingsService;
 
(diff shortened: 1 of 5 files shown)


n13 (commented):
Reviewed specifically for device-locking regressions: traced the full lifecycle against `app_lifecycle_manager.dart`, `auth_wrapper.dart` and `app.dart` at this head, checked every caller of the changed code, and ran the unit tests locally — 19/19 pass (`local_auth_service_test.dart` + `wallet_creation_service_test.dart`), CI analyze green.

**What's solid**

- Fail-closed direction is correct everywhere: in-memory `_lastPausedTime`, null + wallet → locked, `isAuthenticated` defaults to false, and gating the grace window on `state.isAuthenticated` closes the locked-session bypass described in the PR body.
- No double-prompt / prompt-loop paths: the `alreadyAuthenticating` guard in `checkAuthentication` and the `isAuthenticating` guard at `app_lifecycle_manager.dart:117` compose correctly — backgrounding with the prompt up neither records a pause time nor retriggers a check on resume, and the pending `authenticate()` future resolves state. During any pending controller prompt `isAuthenticated` is always false, so the new early-return can't clobber a real prompt's state.
- The async gap in `recordBackgroundTime` (awaits `getHasWallet` before stamping) can only make the grace timestamp fresher, never staler — fails safe.
- Per-action gates (send, multisig, recovery phrase, reset) call `LocalAuthService().authenticate()` directly and are unaffected by the state-machine change.
- SDK cleanup is safe: `getLastPausedTime`/`setLastPausedTime`/`cleanLastPausedTime` and the `last_paused_time` key have no other consumers anywhere in the monorepo at this head (checked miner-app, cold-wallet-app, mobile-app); mock regen matches.

**Two non-blocking concerns**

1. **Fresh-install lock-screen flash (minor UX).** `AuthWrapper` sits in the `MaterialApp.builder` stack above every route (`app.dart:55`), and state now starts locked, so a first launch with no wallet briefly renders the "Authorization Required / Unlock Wallet" screen before `checkAuthentication` → `authenticate()` auto-passes (no wallet → returns `true`). Harmless, but visible on every cold start for pre-wallet users. If you want to avoid it, short-circuit on `!getHasWallet()` in `checkAuthentication` before the locked-session fall-through — that does not reopen the bypass, which needed `shouldRequireAuthentication() == false` with an *existing* wallet.
2. **No-device-security wallets now hit a dead end at cold start.** Consistent with #567's fail-closed posture and called out in the PR body, but worth stating the blast radius: previously a missing timestamp failed open, so these users could still get in via swipe-kill/relaunch; now the lock screen is permanent and its only affordance retries an auth that can never succeed — no path to settings/reset. If that is the intended product posture, fine; otherwise a follow-up escape hatch (e.g. a reset-wallet link on the lock screen) would preempt support tickets.

**Residual from the M1 finding (out of scope here):** the lock is still overlay-only — the unlocked route tree builds and stays alive underneath the `AuthWrapper` scaffold (polling keeps running, semantics tree still exposed). Consider `ExcludeSemantics` or gating the child in a follow-up.

**Verdict: Approve.** The lock-semantics change is deliberate and fails closed; I found no path where a legitimately unlocked session gets stuck or a locked session gets in without a prompt. Agree with the PR body that force-stop / reboot / prompt-oscillation deserve a manual smoke test on both platforms before release — I verified statically and via unit tests, not on hardware.

mobile-app/lib/services/local_auth_service.dart
```diff
@@ -9,15 +9,26 @@ class LocalAuthService {
 
   final LocalAuthentication _localAuth;
   final SettingsService _settingsService;
+  final Duration _authTimeout;
 
-  LocalAuthService._internal() : _localAuth = LocalAuthentication(), _settingsService = SettingsService();
+  /// In-memory only: the background grace window must never survive
+  /// process death. A null value with an existing wallet therefore means
+  /// "this process has not recorded a backgrounding" — i.e. locked.
+  DateTime? _lastPausedTime;
```
dewabisma, line 17: Okay, so if we kill app, it will be locked instantly. Only backgrounding can have grace window.

dewabisma (approved):
LGTM!

PR #570, opened 2026-07-23
fix: gate dev seed import behind kDebugMode (M8)

n13:
## Summary
- Gate `HdWalletService._devAccounts` and `isDevAccount` behind `kDebugMode` in quantus_sdk, so the hardcoded dev keypairs (//Crystal Alice/Bob/Charlie) are only reachable in debug builds.
- Gate the import screen's `//`-prefix word-count validation bypass behind `kDebugMode`. In release builds a `//`-prefixed input now goes through normal 12/24-word mnemonic validation and fails like any invalid phrase.

Addresses finding M8 of the 2026-07-22 mobile wallet security audit.

## Verification
- `dart analyze` clean in both `quantus_sdk/` and `mobile-app/`.
- No existing tests reference the dev-seed path; `kDebugMode` is true in tests, so dev-seed-dependent flows still work under test.
- `flutter test test/generate_keys_test.dart` cannot run in this worktree because the Rust native library is not built here (fails identically on a clean checkout at setUpAll library load — pre-existing environment limitation, unrelated to this change).


n13 (commented):
simple, looks good

PR #581, opened 2026-07-23
fix: pin SS58 prefix to 189 and harden Rust error handling (L5, L6)

n13:
## Summary

Addresses findings **L5** and **L6** of the 2026-07-22 mobile wallet security audit.

**L5 — Foreign SS58 network prefixes accepted as valid recipients**
- `ss58_to_account_id` (`quantus_sdk/rust/src/api/crypto.rs`) now decodes with `AccountId32::from_ss58check_with_version` and rejects any prefix other than the Quantus prefix **189** (verified against the chain runtime's `Ss58AddressFormat::custom(189)` and `AppConstants.ss58prefix`). Previously any registered prefix (Polkadot 0, Kusama 2, generic 42, ...) validated and decoded to an uncontrolled `AccountId32`, burning funds.
- Dart-side `isValidSS58Address` and all other `ss58ToAccountId` call sites keep working unchanged: the FRB Dart signature is still `Uint8List ss58ToAccountId(...)`, it now throws on invalid/foreign input instead of aborting on a Rust panic. Foreign prefixes are now rejected.

**L6 — Rust API panic style on user-controlled input**
- `ss58_to_account_id`: `unwrap` → `Result<Vec<u8>, String>` (matches the existing `Result<T, String>` pattern used throughout `wormhole.rs` / `first_hash_to_address`).
- `generate_keypair`: `expect` → `Result<Keypair, HDLatticeError>` (same error type already used by `generate_derived_keypair` / `derive_wormhole`).
- `wormhole_compute_output_amount`: checked arithmetic (`checked_sub` for the `fee_bps > 10000` underflow, `checked_mul`, `u32::try_from`) → `Result<u32, String>`.
- `decode_leaf_amount`: silent `as u32` truncation → checked `u32::try_from` with an error. Kept the `u32` return type (consistent with `ProofInput` amount fields) so the Dart signature is unchanged.
- Removed the dead, panicking `derive_hd_path` FRB export — verified zero Dart callers outside the generated bindings (grepped `quantus_sdk/lib`, `mobile-app/lib`, `miner-app/lib`, `cold-wallet-app/lib`). Its only dependency, `nam-tiny-hderive`, was removed from `Cargo.toml`.

## Bindings regeneration

FRB bindings were regenerated with `flutter_rust_bridge_codegen generate` (v2.12.0, matching the pinned `flutter_rust_bridge = "=2.12.0"`). The public Dart API is source-compatible: all changed functions keep their Dart return types and now throw instead of panicking; all existing call sites already wrap these calls in try/catch or propagate exceptions.

Note: the local Dart formatter (SDK 3.12.0) re-wrapped parts of the regenerated files; purely mechanical formatting churn in `frb_generated.*`. Formatting-only changes to the untouched `api/multisig.dart` and `api/ur.dart` were reverted to keep the diff minimal.

## Test fixes

- `generate_keys_test.dart`: the wormhole test compared against a generic Substrate (prefix 42) address; replaced with the prefix-189 encoding of the same account bytes, and added a test asserting `ss58ToAccountId` rejects the foreign-prefix and garbage inputs.
- `recovery_proxy_encoding_test.dart`: `setUpAll` now calls `setDefaultSs58Prefix(prefix: 189)` (it only ran `RustLib.init()`, so `toAccountId` encoded with the sp-core default prefix 42, which is now rejected on decode).

## Verification

- `cargo check` — clean
- `cargo clippy` — clean (one pre-existing `needless_range_loop` warning in untouched `generate_proof` code)
- `cargo test --lib` — 13/13 passed
- `dart analyze` (quantus_sdk) — no issues
- `flutter test` (quantus_sdk, full suite, incl. native tests against a release build of the Rust lib) — 134/134 passed

## Scope notes

- `rust-transaction-parser` was inspected per the audit scope and left alone — no SS58 decoding there.
- Other `expect`s on internally-generated key material (`to_ml_dsa`, `generate_keypair_from_seed`, `sign_message`) were left as-is; the audit flags only the user-controlled entry points fixed here.


n13 (commented):
LGTM

PR #582, opened 2026-07-23
fix: queue intents until unlocked and bind send to source account (M10, L10)

n13:
## Summary

Addresses findings M10 and L10 of the 2026-07-22 mobile wallet security audit, plus the review findings from the first round.

### M10 — intents consumed beneath the lock overlay

Deep-link (`/pay`, `/account`) and notification (transaction, multisig proposal) intents were acted on by `HomeScreen`'s listeners while the lock overlay was up.

- All intent handlers in `home_screen.dart` require `localAuthProvider.isAuthenticated` before consuming.
- Intents arriving while locked stay queued in their `StateProvider`s and are drained when `isAuthenticated` flips to true. The cold-start retry on `activeAccountProvider` drains all pending intents.

### L10 — source account re-resolved at submit time

`RegularSendStrategy` re-resolved the signing account at submit time, so an account switch between review and confirm signed from the new account.

- `RegularSendStrategy` captures the source `Account` at flow start (matching the other strategies) and uses it for fee estimation, balance validation and submission.
- `effectiveMaxBalanceProviderFamily(accountId)` binds spendable-balance validation to the captured account.

### Send-flow lifecycle — single entry point, one flow at a time

All send flows start through `startSendFlow()` (`send_providers.dart`), the only place that mutates send-flow state:

- Sets `sendFlowActiveProvider` (bool) before pushing and clears it by awaiting the route's `popped` future — completed on pop, replacement, and `pushAndRemoveUntil`, so the flag is correct however the flow exits. No provider mutation in widget `initState`/`dispose`, no post-frame callbacks.
- Refuses to start a second flow while one is active (also guards double-tap).
- Owns the Keystone signing-session reset (`startNewSendSession`), so entry points no longer touch the keystone cache; a QR cached by an earlier flow (potentially stale nonce) can never be reused.
- Intents that would start or interrupt a send (payment, shared-account, proposal) arriving while a send is in flight are consumed and dropped with a log line — never queued to fire later.

### First-round review findings

1. Balance-error fallback: `effectiveBalanceProviderFamily` keeps the last fetched balance on refresh errors instead of zeroing spendable-balance checks; errors before any successful fetch still propagate.
2. Non-regular active account: payment intents and the shared-address sheet's Send tap surface a "Switch to a regular account to send" warning toast (new l10n string, en + id) instead of failing silently; the sheet stays open.
3. `_isUnlocked` documents the deliberate choice to gate on authentication only, not the visual privacy overlay.
4. Tests added (see below).
5. All logging goes through `quantusPrint`.
6. Interleaved intents resolved structurally: the in-flight flag is set synchronously at entry, so a second queued intent in the same drain pass is dropped.

### Merge with main

Merged current main (through #580). Conflict in `shared_address_action_sheet.dart` resolved to run main's SS58 fail-closed check first, then the regular-account guard, then `startSendFlow`.

## Verification

- `flutter test test/unit test/screens`: 266/266 passed.
- New tests: intent queued while locked drains on unlock; payment intent opens the flow bound to the captured account and resets the Keystone session; intents arriving mid-send are dropped; shared-address sheet warns and stays open for non-regular accounts; effective balance keeps last-good value on refresh error and propagates first-fetch errors; Keystone cache fixtures deduplicated into `test/fakes.dart`.

## Notes

- The multisig-propose and encrypted flows already carried their source account in the strategy; only the regular send needed the L10 change.
- Transaction-detail intents are gated on unlock but not dropped mid-send (informational sheet only, no account switch).


mobile-app/lib/features/components/shared_address_action_sheet.dart
```diff
@@ -60,12 +61,21 @@ class _SharedAddressActionSheetState extends State<SharedAddressActionSheet> {
...
 
   void _sendToAddress() {
-    ProviderScope.containerOf(context).read(keystoneSignCacheProvider.notifier).startNewSendSession();
+    final container = ProviderScope.containerOf(context);
+    final active = container.read(activeAccountProvider).value;
+    if (active is! RegularAccount) {
+      debugPrint('shared address send: active account cannot send regular transfers');
+      return;
+    }
+    container.read(keystoneSignCacheProvider.notifier).startNewSendSession();
```
n13: why does a normal send involve the keystone provider?

n13 (commented):
## Review

Checked out the branch locally, ran `flutter analyze` and the new/changed test files — all 21 tests pass, and the logic holds up under reading. The fixes for M10 (intent gating behind the lock overlay) and L10 (source account captured at flow start) address the audit findings correctly. The `startSendFlow()` single-entry-point refactor is a genuine improvement: no provider mutation in widget lifecycle, the double-start guard is race-free (read+set with no await in between), and awaiting the route's popped future covers pop/replacement/removal. Test coverage of the lock-queue/drain, in-flight-drop, and account-switch scenarios is exactly the right shape.

### Blocking: CI is red

The `Analyze` check fails on the format step. These four files are not `dart format --line-length=120` clean:

- `lib/features/components/shared_address_action_sheet.dart`
- `lib/providers/wallet_providers.dart`
- `lib/v2/screens/home/home_screen.dart`
- `test/screens/home_intent_gating_test.dart`

`dart format lib test --line-length=120` in `mobile-app` fixes it.

### Should fix: unused imports

`flutter analyze` reports two unused imports added by this PR:

- `lib/v2/screens/send/input_amount_screen.dart:16` — `send_providers.dart`
- `lib/v2/screens/send/select_recipient_screen.dart:25` — `send_providers.dart`

### Non-blocking observations

- `shared_address_action_sheet.dart:75` — the `active is! RegularAccount` check conflates "accounts still loading" (`value == null`) with "wrong account type". On a cold start, a fast tap on "Send To This Account" would show "Switch to a regular account to send" when really nothing is loaded yet. Very unlikely in practice (home screen gates on account load), but a `null → return early` branch would make it precise.
- `_onTransactionIntent` is deliberately not gated on `sendFlowActiveProvider`, so a transaction notification can open a detail sheet over a send flow. That's safe now that the source account is captured, and it neither switches accounts nor starts a send — assuming this is intentional, fine; a one-line comment saying so would help.
- The stale-balance-on-refresh-error change in `effectiveBalanceProviderFamily` is a good trade for spendable-balance checks mid-send; errors before any successful fetch still propagate, and the test pins both paths.
- Dropping (rather than queuing) payment/shared/proposal intents during an active send is the right call for a security fix; it's logged via `quantusPrint`, so support can diagnose "my deep link did nothing".

## Verdict: Request changes — trivial

The substance is correct and well-tested. The only blockers are mechanical: run `dart format --line-length=120` and delete the two unused imports so CI goes green. The null-vs-wrong-type edge in the shared-address sheet is worth a follow-up but shouldn't hold the merge.

*(Posted as a comment because GitHub doesn't allow requesting changes on your own PR.)*

n13 (commented):
## Review verdict: ❌ Request changes

The L10 account capture is complete and the targeted tests pass, but this head is not ready to merge because M10 still has a resume-time lock race and CI has two independent failures.

### Blocking findings

1. **M10 can still consume intents while the app is visually locked during resume.**

   HomeScreen._isUnlocked checks only localAuthProvider.isAuthenticated. After the app has been backgrounded, LocalAuthController can be in the state isAuthenticated == true and isVisuallyLocked == true. On resume, AppLifecycleManager calls checkAuthentication without awaiting it, and checkAuthentication awaits shouldRequireAuthentication before it clears isAuthenticated. An app-link or notification callback delivered in that window passes the current gate, clears its intent provider, and navigates beneath the overlay; only afterward can authentication flip to false. This recreates the behavior M10 is intended to remove for background sessions that require re-authentication.

   Please block intent consumption while isVisuallyLocked is true (or expose an equivalent explicit auth-check-pending state), and drain when the visual lock clears even when isAuthenticated stays true for a grace-period resume. Add a test for the isAuthenticated == true / isVisuallyLocked == true state transitioning to unlocked; the current test only covers isAuthenticated == false.

2. **The PR does not pass repository checks.**

   The GitHub Analyze job fails formatting. The exact local check reports four files requiring dart format:

   - lib/features/components/shared_address_action_sheet.dart
   - lib/providers/wallet_providers.dart
   - lib/v2/screens/home/home_screen.dart
   - test/screens/home_intent_gating_test.dart

   After that step, flutter analyze also fails on unused send_providers.dart imports in input_amount_screen.dart:16 and select_recipient_screen.dart:25. Remove those imports and rerun formatting/analyze.

### Verification

Reviewed head 8d7b57c9. The five directly affected test files pass locally: 21/21 tests.

PR #585, opened 2026-07-25
fix: multisig approve must carry the proposal's inner call

n13:
## Why

The generated planck metadata was last regenerated in #509 and had drifted from the live chain (**spec 136 / tx version 3**). Regenerating surfaced a live bug and two API changes.

### 1. `Multisig.approve` gained a third field: `call: Vec<u8>`

The runtime's own rustdoc says why:

> The approver must resubmit the proposal's inner call bytes; the approval is only valid if they are byte-equal to the payload stored at `proposal_id`. This binds the approver's signature to the actual call being approved, so offline/cold-wallet signers can decode and inspect what they are signing instead of trusting an opaque proposal id.

The hot wallet was still encoding two fields, so **every approval it built was invalid against the live runtime**. It compiled only because the metadata was stale.

Approve now reads the stored call from **chain storage** (`multisig.proposals`) — the authoritative source the chain byte-compares against — and blocks confirmation until it loads, so an approval is never built from a guess. The indexer's `call_raw` is retained for display only, never as a signing input.

### 2. Scheduler (8), Referenda (10) and ConvictionVoting (12) no longer expose dispatchables

Governance is now Preimage (7), TechCollective (13) and TechReferenda (14). No hand-written code referenced the removed pallets, so their generated files are deleted.

### 3. ReversibleTransfers gained asset-transfer variants

`schedule_asset_transfer` / `schedule_asset_transfer_with_delay`.

## Proposal display fix

The confirm sheets and proposal rows hardcoded a transfer shape — they rendered `proposal.amount` and `proposal.recipient` unconditionally, so a governance vote or upgrade authorisation displayed as **"0 QUAN to ''"**.

This adds a shared call decoder (`quantus_sdk/lib/src/chain/`) that turns any runtime call into a display tree carrying **every** parameter. It is built on the generated polkadart codecs, so coverage follows the metadata instead of a hand-written pallet switch, and an unknown variant throws rather than being silently skipped. Calls without a bespoke describer fall back to a generic walk of the generated `toJson()`, so no field is ever hidden.

Proposal display now shows the call the runtime actually declared with its parameters, and lifts a nested transfer's amount into the hero position — so an approval still reads as "1 QUAN to qz…" while a vote reads as "TechCollective · vote".

Describers cover transfers (native, asset, reversible), all 7 multisig calls, governance (TechCollective / TechReferenda / Preimage / System upgrades), treasury, utility batches and recovery. Nested calls decode recursively and strictly: trailing bytes or an undecodable inner call are rejected rather than rendered as hex.

### 4. Cold wallet logo changed
Self explanatory, somehow added this to this branch b/c I was fixing the cold wallet and hot wallet simultaneously 

### 5. NPE fixed found during testing

fixes a null pointer exception in rendering under certain conditions


mobile-app/lib/providers/account_providers.dart
```diff
@@ -45,7 +46,7 @@ class AccountsNotifier extends StateNotifier<AsyncValue<List<Account>>> {
   }
 
   Account? getAccountWithId(String accountId) {
-    return state.value?.firstWhere((account) => account.accountId == accountId);
+    return state.value?.firstWhereOrNull((account) => account.accountId == accountId);
```
n13, line 49: fixes a null pointer exception in rendering under certain conditions

n13 (commented):
## Review — Opus 4.6

### Summary

Large, high-impact PR that fixes a live correctness bug (multisig approve was missing the third `call` field), regenerates chain metadata from spec 129→136 / tx version 2→3, adds a shared call decoder in `quantus_sdk`, and fixes proposal display for non-transfer calls. The net line count is -3566 which is mostly dead generated code removal.

### Findings

**Correctness (Critical fix — good)**

- The approve extrinsic was silently building invalid payloads due to stale metadata. The fix is clearly correct: read the authoritative bytes from chain storage (`fetchProposalCallBytes`), block the confirm sheet until they load, and pass them through to both the fee estimate and the submitted extrinsic. This is exactly the right design.

**Architecture (Good)**

- `CallDecoder.describe()` + `DecodedCall` + `CallField` sealed hierarchy is clean and extensible. The fallback via `toJson()` walk means no call is ever silently dropped — good fail-loud principle.
- The `MultisigConfirmCallBytesLoader` typedef pattern on the confirm sheet is sensible. Callers that don't need it (cancel, execute) pass `null` and their type signatures carry `callBytes` through untouched.

**Minor observations (non-blocking)**

1. **`_fallbackTransferCall` in the confirm sheet** — constructs a `DecodedCall` with `fields: const []` when neither the chain bytes nor the indexer's decoded call are available. This means the sheet renders just the headline, which is fine, but a `DecodedCallView` for an empty-fields call is a no-op widget. Cosmetic only, no action needed.

2. **NPE fix (`firstWhereOrNull`)** in `account_providers.dart` — the old `firstWhere` with no `orElse` would throw if the account list didn't contain the id. Good catch. The `collection` dependency was already present.

3. **`ProposalData.callWeight` removal** — the generated type no longer carries `callWeight`. This is a backwards-incompatible storage change at the codec level. Confirmed this tracks the runtime's actual storage struct at spec 136, so it's correct.

4. **`ITSAppUsesNonExemptEncryption: false`** in the cold wallet's Info.plist — correctly declares no non-exempt encryption. Saves the daily "missing export compliance" prompt during App Store Connect upload.

5. **`upload_ios_build.sh`** — uses `flutter clean` which is slow but safe for release builds. The `ls ... | head -n 1` glob works because this project always produces exactly one IPA. Fine for a helper script.

### Verdict

**Approve.** The core fix is clearly correct, well-tested by the rewritten `quantus_payload_parser_test.dart`, and the design properly separates the authoritative on-chain bytes (for signing) from the indexer's copy (for display-only). The metadata regen is a necessary mechanical diff. No blocking issues found.

---
*Reviewed by Opus 4.6*

dewabisma (approved):
LGTM!

PR #586, opened 2026-07-25
feat: cold wallet decodes every chain call, no blind signing

n13:
> **Stacked on #585** — base is `fix/multisig-approve-with-call`. Review that one first; this PR's diff is only the cold-wallet layer.

## Why

The cold wallet could decode **two pallets**. `QuantusPayloadParser` carried a hand-written pallet-index switch for Balances and ReversibleTransfers, so every other call hard-failed with `Unknown pallet index` — the core dev team could not use a cold wallet for multisig approvals or governance at all.

## Parser

The call now decodes through the generated polkadart codecs (`CallDecoder`, added in #585), so coverage follows the metadata instead of an allowlist that falls behind the runtime.

Every existing hardening check stays: 8 KB cap, strict whole-payload decode, zero trailing bytes, era/nonce/tip/metadata-mode bounds, mode↔hash consistency, known genesis hash. An inner call that does not decode exactly **rejects the whole payload** rather than appearing as an opaque blob next to a Sign button.

`ParsedPayload` now carries the `DecodedCall` tree, the raw bytes, and `specMatchesBundled`.

## Review screen

Rebuilt around showing everything that gets signed:

- **Amount hero** comes from the call's transfer summary, which propagates up from a nested call — a multisig approve of a transfer still reads as an amount and a recipient, while a governance vote names the call instead of claiming zero.
- **Every parameter** in signed order; nested calls as boxed sub-lists, recursively. Addresses get a checkphrase — all of them, not just a destination. Hashes and byte blobs shown in full, wrapped, never truncated.
- **Signing key** address + checkphrase: the payload does not name the signer, and which key signs is part of what is happening.
- **Signed extensions** (network, runtime versions, nonce, era, tip, genesis / block / metadata hashes) plus a collapsible **raw payload hex**.
- **Sign is gated on scrolling** to the end of the parameter list, so "I read it" is an action rather than an assumption. Payloads short enough to fit on one screen release the gate after layout.
- **Spec-drift warning** instead of silent acceptance. This is not theoretical: the parser test covers a retired-devnet vector whose pallet index 13 was ReversibleTransfers and is TechCollective on this runtime — those bytes now decode as a plausible-looking `vote` and only fail when the extensions don't parse. Across runtime versions an index can mean a different call, so a mismatched spec gets a red banner naming both versions. Per the agreed policy this warns rather than blocks.

## Debug buttons

Three `kDebugMode`-only buttons on the scanner (send, multisig approve wrapping a transfer, governance aye vote) inject a payload directly, so the scan → review → sign path can be exercised on a simulator, which has no camera. They share one extension encoder, covered by a test so they can't rot silently into "could not read transaction".

## Testing

- `melos analyze` clean across all four packages.
- `quantus_sdk`: 160 tests. `quantus_payload_parser_test.dart` rewritten on the new model — every prior hardening case kept, plus multisig approve/propose, `TechCollective.vote`, `System.authorize_upgrade`, `note_preimage` wrapping an upgrade, spec-drift, and an undecodable inner call.
- `cold-wallet-app`: 7 tests, including the byte-exact hot→cold transfer vector and an approve-with-inner-transfer end to end.
- **Ran on the iOS simulator** and reviewed the rendered screen: hero showed `4.2 QUAN` lifted from the transfer nested inside a `Multisig · approve`, both the multisig account and the inner destination resolved checkphrases offline, extensions and raw-payload sections rendered, and Sign correctly stayed disabled until the page had been read. Running it also caught a real crash — `maxScrollExtent` read during `build` before the scroll position had content dimensions — now fixed and guarded.

Not yet exercised against a live chain end to end (hot wallet proposes → cold wallet approves → broadcast); worth doing with #585 before merge.

### DIFF
--- a/cold-wallet-app/lib/components/call_detail_view.dart
+++ b/cold-wallet-app/lib/components/call_detail_view.dart
@@ -1,0 +1,162 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:quantus_sdk/quantus_sdk.dart';
+import 'package:quantus_cold_wallet/components/detail_row.dart';
+import 'package:quantus_cold_wallet/theme/app_colors.dart';
+import 'package:quantus_cold_wallet/theme/app_text_styles.dart';
+
+/// Resolves human checkphrases for every address on screen, not just a
+/// destination — an approval or a governance call can name several accounts, and
+/// each one needs to be verifiable by eye.
+final _checkphraseProvider = FutureProvider.family<String, String>((ref, address) async {
+  return HumanReadableChecksumService().getHumanReadableName(address);
+});
+
+/// Renders every parameter of a decoded call, recursing into nested calls.
+///
+/// This is the whole point of the signing screen: nothing is summarised away and
+/// nothing is hidden behind a "details" tap. If a field is in the bytes being
+/// signed, it is on this screen.
+class CallDetailView extends StatelessWidget {
+  final DecodedCall call;
+
+  /// Nesting depth; nested calls are boxed and indented.
+  final int depth;
+
+  const CallDetailView({super.key, required this.call, this.depth = 0});
+
+  @override
+  Widget build(BuildContext context) {
+    final colors = context.colors;
+    final text = context.themeText;
+
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.stretch,
+      children: [
+        Text(
+          call.displayTitle,
+          style: text.transactionDetailRowValue?.copyWith(
+            color: depth == 0 ? colors.textPrimary : colors.checksum,
+            fontWeight: FontWeight.w600,
+          ),
+        ),
+        for (final field in call.fields) _CallFieldView(field: field, depth: depth),
+      ],
+    );
+  }
+}
+
+class _CallFieldView extends ConsumerWidget {
+  final CallField field;
+  final int depth;
+
+  const _CallFieldView({required this.field, required this.depth});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final colors = context.colors;
+    final text = context.themeText;
+
+    switch (field) {
+      case ValueField(:final label, :final value, :final kind, :final note):
+        if (kind == ValueKind.address) {
+          return _AddressField(label: label, address: value, note: note);
+        }
+        return DetailRow(
+          label: label,
+          value: value,
+          monospace: kind == ValueKind.hash || kind == ValueKind.bytes,
+          note: note,
+        );
+
+      case AmountField(:final label, :final planck, :final assetId, :final note):
+        return DetailRow(
+          label: label,
+          value: assetId == null
+              ? '${_formatPlanck(planck)} ${AppConstants.tokenSymbol}'
+              : '$planck raw units of asset #$assetId',
+          note: assetId == null
+              ? note
+              : [
+                  note,
+                  'Asset decimals are not part of this payload, so the raw amount is shown.',
+                ].whereType<String>().join(' '),
+        );
+
+      case FieldGroup(:final label, :final items):
+        return Padding(
+          padding: const EdgeInsets.only(top: 10),
+          child: Column(
+            crossAxisAlignment: CrossAxisAlignment.stretch,
+            children: [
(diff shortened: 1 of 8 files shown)


n13 (commented):
## Review — Opus 4.6

### Summary

Cold wallet no-blind-signing layer, stacked on #585. Replaces the hand-written two-pallet parser with `CallDecoder` so every chain call is decoded and displayed. The review screen is rebuilt to show every parameter in signed order, with a scroll-to-sign gate. +904/-271 across 8 files.

### Findings

**Design (Excellent)**

- **Fail closed**: an inner call that doesn't decode exactly rejects the whole payload. No "sign this opaque blob" path exists. This is the correct security posture for a cold wallet.
- **Spec-drift warning** instead of hard block: across runtime versions the same pallet index can mean a different call (demonstrated concretely by the retired-devnet test vector where index 13 was ReversibleTransfers but is now TechCollective). Showing a red banner naming both spec versions is the right middle ground.
- **Scroll-to-sign gate**: `_reviewedToEnd` tracks whether the user has scrolled to the bottom. Short payloads that fit on one screen release the gate after layout via `addPostFrameCallback`. This avoids the "read before you sign" problem while not breaking UX for trivial transfers.

**Implementation (Good)**

- `CallDetailView` recurses through `NestedCallField` with a boxed container for depth, making nested multisig approvals visually distinct. Every address gets a checkphrase via `_checkphraseProvider` — including multisig account addresses, not just destinations.
- `DebugPayloads` shares one extension encoder (`_withExtensions`) with the test, so they can't silently rot — the test asserts `specMatchesBundled` is true, which fails if the extensions diverge from the bundled metadata constants.
- The `ParsedPayload` now carries `raw` bytes and `specMatchesBundled`, clean additions.

**Minor observations (non-blocking)**

1. **`_onScroll` tolerance** — the check `position.pixels >= position.maxScrollExtent - 16` uses a 16px tolerance. This is fine on iOS but might be tight on some Android devices with imprecise flings. If any QA reports "can't sign", bumping to 24px would be safe. Not blocking.

2. **`_scheduleReviewGateCheck` runs every build** — when the widget tree rebuilds (e.g. setState from error display), this schedules another postFrameCallback. It's idempotent (early-returns if `_reviewedToEnd` is already true) but slightly noisy. Fine as-is.

3. **Raw payload hex section** — shows the full payload bytes in a collapsible section. Good for advanced users who want to verify against another tool.

4. **Test coverage** — the `DebugPayloads` group validates the governance vote path parses with no spec drift, and the `transaction_signing_test.dart` covers the multisig-approve-with-transfer end-to-end. The cold wallet can't easily test the UI in a headless unit test, but the parser coverage in `quantus_sdk` is comprehensive.

5. **`DetailRow` as a shared component** — extracts the inline `_detailRow` method into a proper widget. Good refactoring.

### Not yet tested (noted in PR)

Hot wallet proposes → cold wallet approves → broadcast against a live chain. Worth doing before merge, as the PR notes.

### Verdict

**Approve.** The security model is sound: every byte that gets signed is displayed, undecodable payloads are hard-rejected, spec drift is flagged, and the sign button is gated on having read the full content. No blocking issues.

---
*Reviewed by Opus 4.6*

cold-wallet-app/lib/components/call_detail_view.dart
```diff
@@ -0,0 +1,162 @@
...
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:quantus_sdk/quantus_sdk.dart';
+import 'package:quantus_cold_wallet/components/detail_row.dart';
+import 'package:quantus_cold_wallet/theme/app_colors.dart';
+import 'package:quantus_cold_wallet/theme/app_text_styles.dart';
+
+/// Resolves human checkphrases for every address on screen, not just a
+/// destination — an approval or a governance call can name several accounts, and
+/// each one needs to be verifiable by eye.
+final _checkphraseProvider = FutureProvider.family<String, String>((ref, address) async {
```
dewabisma, line 11: We can add autoDispose so preventing memory bloated by unused checkphrases cache if that is reasonable.

dewabisma (approved):
LGTM! Just one suggestion if that is reasonable to have

cold-wallet-app/lib/components/call_detail_view.dart
n13: I don't think we need to you'd have to have thousands cached to even notice, or more .... 10s of thousands...

PR #587, opened 2026-07-27
Cold wallet improvements

n13:
## Summary

Cold wallet UX improvements focused on the import and security flows:

- **Reimplemented import wallet screen** — replaced the word-by-word BIP-39 entry with a single multiline paste field. Uses a custom `_ObscuringMnemonicController` to mask the mnemonic on screen (rendered as `x` characters, with an eye toggle to reveal) while preserving the real text for import. Validation is a 12/24-word count check plus key derivation via `HdWalletService().keyPairAtIndex`, which throws on an invalid phrase; `//` dev phrases bypass the word count check.

- **Removed password restrictions** — no minimum length enforced, empty password allowed on both the set-password and lock screens. The cold wallet is air-gapped; local security policy is the user's call, not ours.

- **Connectivity guard disabled in debug builds** — `ConnectivityGuard` now returns early when `kDebugMode`, so the air-gap blocker doesn't fire during development. Release builds are unchanged (still fail closed).

- **Removed stale doc comments** in `sign_transaction_screen.dart` and `quantus_payload_parser.dart`.

- **Version bump** to 1.0.1+2.

- **iOS project cleanup** — removed the `Pods` group display-name entry from `project.pbxproj`.

## Test plan

- [x] Import a 12-word and 24-word mnemonic via paste on the new import screen
- [x] Verify mnemonic is obscured during entry and correct wallet is derived
- [x] Set a short (< 8 char) password and confirm it works on the lock screen
- [x] Set an empty password and confirm unlock still works
- [x] Cold-sign a transfer and a multisig approval end-to-end via QR


n13 (commented):
## Review

Checked the diff against the repo, including the Rust derivation path the new import flow relies on.

### What's good

- The paste-field rework is a real UX improvement over word-by-word entry, and the `_ObscuringMnemonicController` approach (equal-length mask so caret/selection stay accurate) is the standard workaround for multiline `obscureText`.
- Validation is not weakened: `keyPairAtIndex` → `derive_key_from_mnemonic` → `Mnemonic::parse_in_normalized` in `qp-rusty-crystals-hdwallet`, which checks wordlist membership *and* checksum — equivalent to the old `Mnemonic.fromSentence`.
- Empty-password flow is technically coherent: Argon2id + AES-GCM work fine with an empty secret, and biometric unlock still stores real key bytes.

### Issues

**1. PR description contradicts the diff**

- "Connectivity guard — added null-safety check" — the actual change is `if (kDebugMode) return const SizedBox.shrink();`, which disables the air-gap blocker entirely in debug builds (`cold-wallet-app/lib/widgets/connectivity_guard.dart:19`). Reasonable dev convenience, but it's not a null-safety fix, and it bypasses the documented fail-closed behavior in debug.
- "The sign transaction screen and payload parser have no net changes" — the diff deletes five doc comments in exactly those files. Two of them carry real security context: the `_specDriftBanner` note ("labels below may belong to a different call than the one that will execute") and the `_decodeCall` note about over/under-reads surfacing as malformed extensions. Please restore these — deleting them is a docs regression, not a merge artifact.
- The iOS change is described as "removed stale Xcode build phase entry" but it's just the `name = Pods;` group display-name line. Harmless, mis-described.

**2. Whitespace handling regression in import validation** — `import_wallet_screen.dart` splits on `' '` only. A phrase pasted with newlines (e.g. a backup file with line breaks every 4 words) or tabs will fail with a confusing error. The old code used `RegExp(r'\s+')`. Suggest:

```dart
final words = mnemonic.trim().split(RegExp(r'\s+')).where((w) => w.isNotEmpty).toList();
```

and pass `words.join(' ')` to `SetPasswordScreen`, so the stored mnemonic is normalized rather than whatever raw whitespace was pasted.

**3. Case sensitivity** — the old flow lowercased every word; the new one doesn't. BIP-39 parsing is case-sensitive, so "Apple banana …" fails. Please lowercase the input before validating, and strip the `"Exception: "` prefix from `e.toString()` before showing it to the user.

**4. Sync Rust call blocks the spinner (minor)** — `generate_derived_keypair` is `frb(sync)`, so PBKDF2 + ML-DSA keygen runs on the main isolate. `_isLoading` is set, but no frame renders while the call blocks, so the spinner never animates. Probably fast enough in practice; just noting the loading state is cosmetic.

**5. Empty password (intentional, but worth a warning)** — with an empty password the vault at rest is encrypted under a publicly derivable key, i.e. obfuscation only. Defensible on an air-gapped device and it's the stated intent, but a one-line "no password = anyone holding this device can unlock" hint when the field is empty would be honest UX.

### Nits

- The mask reveals per-word lengths (runs of `x` separated by spaces) — negligible here given the reveal toggle exists.
- The test-file diff is a blank-line removal, pure noise.
- No test covers the new 12/24-word validation — would be a cheap unit test.

### Verdict

The two things I'd ask to change before merge: restore the deleted doc comments in `sign_transaction_screen.dart` / `quantus_payload_parser.dart`, and fix the whitespace splitting in `_import`. Everything else is minor.

dewabisma (approved):
LGTM!

PR #593, opened 2026-08-02
refactor: centralize mnemonic generation in the SDK

n13:
## Why

Seed generation lived in two places: `SubstrateService.generateMnemonic()` (used by the mobile and cold-wallet apps) and a duplicated copy in `MinerWalletService`. Duplicated entropy-generation code is a maintenance and audit risk — a future change to one path could silently diverge from the other.

## What

- `MinerWalletService.generateMnemonic()` now delegates to `SubstrateService().generateMnemonic()`, so all mnemonic/seed generation lives in one spot in `quantus_sdk`.
- The SDK generator now reuses a single `Random.secure()` instance instead of constructing one per entropy byte (no behavior change — each instance was freshly OS-seeded anyway).

Both paths were and remain OS-CSPRNG-backed (`Random.secure()` → SecRandomCopyBytes / getrandom / BCryptGenRandom); this is a consolidation, not a security fix.

## Verification

- `melos run format` — clean
- `melos run analyze` — clean for quantus_sdk, miner-app, cold-wallet-app (mobile-app's analysis server crashes with SIGKILL on this machine, unrelated to this change; mobile-app is untouched)

### DIFF
--- a/miner-app/lib/features/setup/rewards_address_setup_screen.dart
+++ b/miner-app/lib/features/setup/rewards_address_setup_screen.dart
@@ -73,9 +73,10 @@ class _RewardsAddressSetupScreenState extends State<RewardsAddressSetupScreen> {
   }
 
   /// Generate a new 24-word mnemonic.
-  void _generateNewMnemonic() {
+  Future<void> _generateNewMnemonic() async {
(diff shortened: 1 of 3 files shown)


dewabisma (approved):
Cool! This is great!

PR #594, opened 2026-08-02
Accounts management redesign: full-screen accounts, wallet names, import under Advanced

n13:
Implements the updated account management design (Figma node 4184-16283), built on top of the keystone design branch.

## Changes

- **Full-screen Accounts screen** — the bottom-sheet popup is replaced by a full screen (`AccountsScreen`) that slides in from the right with a back button. All add/import/disconnect flows still pop back to it by route name.
- **Main wallet on top, others collapsible** — the first wallet renders expanded with its name, an edit pencil, and an `ACTIVE WALLET` badge. Additional wallets (software and keystone) collapse into a single row showing the wallet's total balance and account count, expanding in place to reveal accounts and the encrypted account. The `Wallet 1 / Wallet 2` segmentation headers are gone.
- **Optional wallet names** — minimal storage in `SettingsService` (`wallet_name_<index>`, cleaned up on `removeWallet`). New `WalletNameScreen` (mirrors the account naming screen) opens from the pencil next to a wallet name; unnamed wallets fall back to "Wallet {n}". The wallet picker in settings also shows custom names.
- **Import flow names the wallet** — after importing a wallet from a recovery phrase, the Wallet Name screen is shown before returning to Accounts with the new account highlighted. Leaving it empty skips naming.
- **Import Account moved under Advanced** — the Add Account screen now shows Add Transparent Account and Add Keystone Wallet as primary options; Advanced (always visible) holds the multisig options (feature-flagged, as before) and Import Account.

## Notes

- `accounts_grouping.dart` reworked from a flat segmented item list to `WalletGroup`s; unit tests updated and extended (`softwareWalletNumber` unchanged for the disconnect-wallet copy).
- Wallet containing the active account (or a newly added highlight target) auto-expands on open.
- Unused sub-segment l10n keys removed; new strings added in English and Indonesian.

## Testing

- `melos run analyze` (--fatal-infos) clean across packages
- mobile-app unit/screen/model tests and quantus_sdk service tests pass
- Ready for simulator testing


mobile-app/lib/v2/screens/accounts/accounts_screen.dart
```diff
@@ -0,0 +1,581 @@
...
+    required AsyncValue<List<MultisigAccount>> multisigAsync,
+    required AsyncValue<DisplayAccount?> activeDisplayAccountAsync,
+    required String? activeAccountId,
+  }) {
+    if (accountsAsync.isLoading || activeDisplayAccountAsync.isLoading || multisigAsync.isLoading) {
+      return const Center(child: Loader());
+    }
+
+    if (accountsAsync.hasError || multisigAsync.hasError) {
+      return Center(
```
dewabisma: This is violating DRY principle, we shouldn't repeat this piece. Only the text content changing.

mobile-app/lib/v2/screens/accounts/accounts_screen.dart
```diff
@@ -0,0 +1,581 @@
...
+              Text(
+                displayName,
+                style: context.themeText.paragraph!.copyWith(fontSize: 18, height: 1),
+                maxLines: 1,
+                overflow: TextOverflow.ellipsis,
+              ),
+              const SizedBox(height: 8),
+              Text(
+                _walletBalanceText(l10n, wallet),
+                style: context.themeText.smallParagraph!.copyWith(fontSize: 14, color: colors.textTertiary, height: 1),
```
dewabisma: I want to nit about small paragraph already fontSize 14 being overridden with value 14 again. Sry about this, but just ridiculous to override fontSize with same value. Might as well not use the textStyle token.

mobile-app/lib/v2/screens/accounts/accounts_screen.dart
```diff
@@ -0,0 +1,581 @@
...
+              ),
+            ],
+          ),
+        ),
+        const SizedBox(width: 8),
+        Text(
+          l10n.accountsScreenAccountCount(wallet.accountCount),
+          style: context.themeText.smallParagraph?.copyWith(fontSize: 14, color: colors.textSubtle, height: 1),
```
dewabisma: Same

mobile-app/lib/v2/screens/accounts/accounts_screen.dart
```diff
@@ -0,0 +1,581 @@
...
+    return Container(
+      padding: const EdgeInsets.symmetric(horizontal: 8, vertical: 4),
+      decoration: BoxDecoration(
+        borderRadius: BorderRadius.circular(4),
+        border: Border.all(color: accent.useOpacity(0.1)),
+        color: accent.useOpacity(0.1),
+      ),
+      child: Text(
+        label.toUpperCase(),
+        style: context.themeText.detail?.copyWith(
```
dewabisma, line 474: We have tiny for font size 11, but anyway, like I mentioned in the other PR, we need to discuss about theme style so we don't keep pilling noise. Because right now already not like the original goal when we add this text theme token.

dewabisma (changes requested):
Few nits.

Overall looking good. 

We really need to discuss about the text theme soon. So, we don't keep dirtying our codebase with the misuse.

mobile-app/lib/v2/screens/accounts/accounts_screen.dart
n13: I think it's basically from the design system

Our current design system is - we have a font type, then various attributes on the fonts

We don't have a system with a few specific types

Could ask Sejal about this

dewabisma (approved):
LGTM!

PR #596, opened 2026-08-05
Fix swap button icon color to match send/receive

n13:
The home screen swap icon SVG has near-white strokes (#F4F6F9) while the send/receive icons are orange (#FF6B35), so the swap button looked different from the other action buttons.

All three action icons now get a `ColorFilter` tint from the theme's `accentOrange` (which is exactly #FF6B35), so send/receive render unchanged, swap matches them, and the icons follow the theme instead of colors baked into the SVGs.


dewabisma (approved):
LGTM!

PR #598, opened 2026-08-08
chore: bump poseidon 3.1.0, plonky2 1.5.5, crystals 4.1.0

n13:
`quantus_sdk/rust` crypto bumps (all crates.io):

| Crate | To |
|-------|-----|
| qp-poseidon-core | **3.1.0** |
| qp-plonky2 | **1.5.5** |
| qp-rusty-crystals-* | **4.1.0** |
| qp-wormhole-* / qp-zk-circuits-common | **4.2.0** |

No git pins for wormhole. Lockfile has a single version of each new crate.

Also - version management for prover binaries and files generated for zk circuits - delete old versions, versioned new files


n13 (commented):
REQUEST_CHANGES (opinion): the dependency bump is not source-compatible with the SDK and also leaves circuit artifact discovery on the removed 3.1.x file layout.

Findings:

1. **[P1] Update the SDK for the new crypto/wormhole APIs before bumping these versions** (`quantus_sdk/rust/Cargo.toml:12-13`, `quantus_sdk/rust/Cargo.toml:25-31`). `cargo test --locked` fails to compile the native SDK with 13 errors. The 4.x crystals APIs made `Keypair`/`WormholePair` fields private, changed mnemonic generation and signing to mutable/borrowed sensitive buffers, and the 4.2 wormhole API changed `PrivateCircuitInputs.secret` to `Secret`; the current callers still use the 3.x forms. This prevents the Rust library, and therefore native SDK/app builds, from being produced. Please adapt all affected call sites and add a CI check that compiles/tests `quantus_sdk/rust` so future dependency-only bumps cannot pass the Flutter-only job while the native library is broken.

2. **[P1] Align the cached artifact list with qp-wormhole 4.2** (`quantus_sdk/rust/src/api/wormhole.rs:244`, `quantus_sdk/lib/src/services/circuit_manager.dart:36`). Both required-file lists still demand `private_batch_prover.bin`, but 4.2 deliberately no longer emits or reads that file: `PrivateBatchProver::new_from_binaries_dir` rebuilds the aggregation prover and consumes only `common.bin`, `verifier.bin`, `dummy_proof.bin`, and `config.json`. Consequently a freshly generated 4.2 directory never satisfies `all_required_files_exist`; every later send/claim regenerates the expensive circuit set, and Dart status/extraction likewise treats the new set as unavailable. Remove the obsolete requirement in both places, align the remaining list with the 4.2 producer/consumer contract, and cover a second `ensure_circuit_binaries` call reusing the first call's output. Since this app does not build public-batch aggregation, also re-evaluate the now-changed `include_prover=true` argument, which in 4.2 generates an unused dummy private-batch proof rather than a prover binary.

Validation:

- `cargo test --locked` — **failed** with 13 compiler errors (`E0616`, `E0061`, `E0608`, `E0277`, `E0308`).
- `cargo metadata --locked --offline --format-version 1 --no-deps` — passed.
- `cargo tree --locked --offline -e normal -p rust_lib_quantus_wallet --depth 1` — passed and resolved the intended direct versions (Poseidon 3.1.0, Plonky2 1.5.5, crystals 4.1.0, wormhole/common 4.2.0).
- `git diff --check d5fc4d6572ebc2244e95009a33f5da5048935867...e3cfac3b1f6ba9216f24ff938f13089e7a737d95` — passed.
- GitHub `Analyze` — passed, but the workflow does not compile `quantus_sdk/rust`.
- `cargo fmt --check` — failed on formatting in unchanged Rust source files; not introduced by this manifest-only diff.

n13 (commented):
## Summary

Version audit of all Quantus-owned crates in `quantus_sdk/rust/Cargo.toml` against crates.io (queried 2026-08-10) and the `quantus_ur` git tags. Every **crates.io** `qp-*` pin in this PR is already at the published latest, and the lockfile has a single resolved version of each. The PR is still not mergeable: `cargo check --locked` fails with 13 errors (4.x crystals / 4.2 wormhole API breaks), and one Quantus-owned dependency is not on its latest tag.

## Crate version audit

| Crate | In PR (toml / lock) | Latest published | Status |
|-------|---------------------|------------------|--------|
| `qp-poseidon-core` | 3.1.0 | 3.1.0 | latest |
| `qp-plonky2` (+ core/field/verifier) | 1.5.5 | 1.5.5 | latest |
| `qp-rusty-crystals-dilithium` | 4.1.0 | 4.1.0 | latest |
| `qp-rusty-crystals-hdwallet` | 4.1.0 | 4.1.0 | latest |
| `qp-wormhole-circuit` | 4.2.0 | 4.2.0 | latest |
| `qp-wormhole-prover` | 4.2.0 | 4.2.0 | latest |
| `qp-wormhole-aggregator` | 4.2.0 | 4.2.0 | latest |
| `qp-wormhole-inputs` | 4.2.0 | 4.2.0 | latest |
| `qp-wormhole-circuit-builder` | 4.2.0 | 4.2.0 | latest |
| `qp-zk-circuits-common` | 4.2.0 | 4.2.0 | latest |
| `quantus_ur` (git tag) | **1.4.0** | **1.6.0** | **behind** |

- No git pins remain for wormhole (all registry sources) — matches the PR claim.
- Lockfile: single version of each `qp-*` / `quantus_ur` package (no dual-resolution).
- `rust-transaction-parser/Cargo.toml` has no Quantus-owned crates.

## Findings

### [P1] SDK does not compile against the bumped crates

`cargo check --locked` in `quantus_sdk/rust` fails with 13 errors (private fields on `Keypair`/`WormholePair`, `SensitiveBytes32` mutability for ML-DSA, `PrivateCircuitInputs.secret` now `Secret`, etc.). Manifest-only bump is not source-compatible. Adapt call sites in `src/api/crypto.rs` and `src/api/wormhole.rs` (and related) before merging. Already covered in an earlier review; reconfirmed on this checkout.

### [P1] Circuit artifact list still expects 3.1 layout

`private_batch_prover.bin` remains in required-file lists while 4.2 no longer produces/consumes it — causes perpetual regeneration. Align with 4.2 consumer contract when fixing the API break.

### [P2] `quantus_ur` not on latest Quantus tag

`quantus_sdk/rust/Cargo.toml:20` pins `tag = "1.4.0"`; repo latest is `1.6.0` (includes inbound UR validation, fragment-length hardening, and ~7x decode speedups — 8 commits ahead of 1.4.0). This is a Quantus-owned crate referenced in Cargo.toml and should be bumped (or an explicit reason documented if 1.4.0 is intentional).

## Verdict: needs changes

crates.io crypto pins in this PR are correct and current. Block on (1) SDK source updates so the bump builds, (2) 4.2 circuit artifact discovery alignment, and preferably (3) `quantus_ur` → `1.6.0`.

n13 (commented):
APPROVE (opinion): the latest commit resolves the native SDK API/build blockers and corrects artifact reuse on the active claim/send path. I found no blocking issues on `f06aafdb8faf36816e5ddcacd1302ae8208fe236`.

The 4.1 crystals migration now uses the sensitive-buffer/accessor APIs correctly, the 4.2 wormhole migration compiles, and the Rust artifact check no longer requires `private_batch_prover.bin`. The private-batch prover is intentionally rebuilt on demand on-device by `PrivateBatchProver::new_from_binaries_dir`.

Non-blocking follow-up: `quantus_sdk/lib/src/services/circuit_manager.dart:36-46` still lists `private_batch_prover.bin`. `quantus_sdk` is unpublished and neither `checkStatus()` nor `extractCircuitsFromAssets()` has an in-repository caller, so this does not affect the active flow; remove the stale entry and 3.1.x wording as cleanup.

I also retract the earlier `include_prover=true` concern: in 4.2 that flag controls the dummy private-batch proof used for public-batch padding, not construction of the real prover, and 3.1 already generated the same dummy proof when the flag was enabled. This PR does not introduce that proving cost.

Validation:

- `cargo test --locked --offline` — passed (13 tests).
- Cross-version derivation probe (3.0.1 vs 4.1.0) — identical ML-DSA public/secret keys and wormhole address/first-hash/secret for the same mnemonic and paths.
- `cargo tree --locked --offline -e normal -p rust_lib_quantus_wallet --depth 1` — passed and resolved the intended direct versions.
- `git diff --check d5fc4d6572ebc2244e95009a33f5da5048935867...f06aafdb8faf36816e5ddcacd1302ae8208fe236` — passed.
- GitHub `Analyze` — passed.
- `cargo fmt --check` and strict Clippy remain unclean due to existing repository formatting and the base-commit `needless_range_loop` at `wormhole.rs:305`; neither is a blocking regression in this PR.

n13 (commented):
Verdict: REQUEST_CHANGES

Blocking finding:

- [P1] Delete every artifact produced by prior wallet versions (`quantus_sdk/rust/src/api/wormhole.rs:19`). `LEGACY_FILES` omits `dummy_private_batch_proof.bin`, which the shipped 3.1 generator wrote because this app called `generate_all_circuit_binaries(..., true, ...)`, and it also omits the 3.0 names `aggregated_prover.bin`, `aggregated_verifier.bin`, and `aggregated_common.bin`. The new `v4.2.0` directory is generated and consumed correctly, but upgrades leave these heavyweight legacy proving artifacts on device. The current cleanup test seeds filenames from `LEGACY_FILES` itself, so it cannot catch omissions. Add the complete historical 3.0/3.1 manifests and test those manifests independently.

Validated at `f3af25d78974ba63c8195b1f125631824cba6a15`: `cargo test --locked` passed (37 tests), both diff checks passed, and live Analyze succeeded. `cargo fmt --all -- --check` is not clean in the changed Rust sources; strict Clippy also reports a pre-existing `needless_range_loop` at `wormhole.rs:359`.

n13 (commented):
APPROVE (opinion): the latest commit resolves the legacy-artifact cleanup blocker. I found no blocking issues on `77093ae42581d3bb04edfa01f0c14ed253aa7788`.

The active wallet path now cleans the complete flat artifact sets written by the shipped 3.0 and 3.1 implementations, removes older version directories, generates/reuses artifacts under `v4.2.0`, and passes that same versioned directory to aggregation. The regression tests define the historical manifests independently of `LEGACY_FILES`, so future omissions will fail rather than testing the implementation against itself.

Validation:

- `cargo fmt --all -- --check` — passed.
- `cargo test --locked` — passed (39 tests).
- `cargo clippy --locked --all-targets --all-features` — passed with only the previously identified base-code `needless_range_loop` warning.
- `git diff --check` for both the latest commit and the complete PR — passed.
- Repository history scan confirmed that the 3.0/3.1 union covers the pre-versioned artifact names written by prior wallet implementations.
- GitHub `Analyze` workflow — passed, including formatting, analysis, Mobile App tests, and Quantus SDK tests.

No blocking findings remain.

dewabisma (approved):
LGTM

PR #600, opened 2026-08-11
Cold wallet: fix broken network-lock override, redesign password fields, 1.2.0+5

n13:
## Summary

- **Fix the network-lock Override button doing nothing.** `ConnectivityGuard` is mounted above the `Navigator` (in the `MaterialApp.builder` stack so it can cover every route), which means `showDialog` from its context threw — there is no Navigator above it, and even with one the dialog would render underneath the full-screen guard. Tapping Override therefore silently failed. The confirmation is now an **inline step inside the guard overlay itself**: Override → full-screen warning with "Override — I accept the risk" / Cancel, no dialog routes involved. Going offline while confirming resets the step. The session-only override and the persistent red "lock overridden" banner behave as before.
- **Make the network lock identical in debug and release.** The `kDebugMode` bypass that suppressed the guard entirely in debug builds is removed — with the Override setting available there is no need for a separate debug behavior, and debug testing now exercises the exact release flow.
- **Redesign the password entry fields** (`PasswordField`, used by Set Password, Confirm Password, and Unlock). Previously the field collapsed to bare text height inside its container (the app theme uses `isDense` + zero content padding), so the inputs looked squished and inconsistent with buttons. Now: fixed 56 px height with vertically centered text, accent-orange border when focused, a properly padded 46 px tap target for the show/hide toggle whose icon brightens on focus, and sensible keyboard actions (`next` on Password, `done`/submit on Confirm and Unlock).
- **Version bump to 1.2.0+5** plus release-build infrastructure so new builds are reproducible: `key.properties`-driven Android release signing config (falls back to debug signing when the file is absent, e.g. CI), `upload_android.sh` to build the signed AAB and optionally upload to the Play internal track, the public Play upload certificate, a gitignore rule for Play service-account credentials, and a faster iOS upload script (`rm -rf build` instead of `flutter clean`), now anchored to its own directory so it is safe to invoke from anywhere.
- Dependency versions stay on the workspace-canonical list (flutter_riverpod 3.3.1, mobile_scanner 7.2.0, local_auth 3.0.1) — an accidental riverpod 3.4.2 bump that required Dart 3.12 against the app's declared 3.8 floor was reverted in 68f51b66.

## Validation

- `flutter analyze .` — clean; repo-pinned `dart format --line-length=120` — no changes; `git diff --check` — clean.
- `flutter test` — cold wallet suite passing.
- Signed release AAB + APK previously built from this branch at versionCode 4 (CN=Quantus Cold Wallet); rebuild at `1.2.0+5` before distribution.


n13 (commented):
REQUEST_CHANGES (opinion): one blocking dependency-compatibility issue remains.

Finding:

- [P1] Align the advertised Dart SDK floor with the upgraded Riverpod dependency (`cold-wallet-app/pubspec.yaml:8,25`). The manifest still allows Dart 3.8, but `flutter_riverpod 3.4.2`/`riverpod 3.4.2` require Dart 3.12 and this PR's lockfile now records `dart: ">=3.12.0 <4.0.0"`. A developer or build using Dart 3.8-3.11, which the manifest claims is supported, cannot resolve this app. Raise the app's SDK constraint (and any aligned workspace/toolchain declaration) to at least 3.12, or retain a Riverpod version compatible with the advertised range.

Validation at base `f8937d35` / head `4851c783`:

- Inspected the complete diff and the intervening debug/release connectivity-lock commit.
- `dart run melos bootstrap` passed after creating the ignored worktree-only `SourcePackages` directories.
- Workspace format and `flutter analyze . --fatal-infos` checks passed; focused cold-wallet analysis also passed on the final head.
- Cold-wallet `flutter test`: 7/7 passed.
- Both upload scripts passed `sh -n`; OpenSSL parsed the new certificate. `git diff --check` flags only the certificate's CRLF line endings.
- GitHub `Analyze` is green.

Android bundle compilation could not be repeated because the reviewer host has no Android SDK. No other blocking findings were found.

n13 (commented):
REQUEST_CHANGES (opinion): the Riverpod/Dart compatibility blocker is fixed, but one destructive release-script issue remains.

Finding:

- [P1] Anchor the broadened cleanup to the mobile-app directory (`mobile-app/upload_ios_build.sh:14`). The script never changes to its own directory, so `rm -rf build` operates on the caller's current working directory. For example, invoking `./mobile-app/upload_ios_build.sh` from the repository root deletes the repository-root `build/`; invoking it by path from another project deletes that project's `build/`. This PR expands the previous target from `build/ios/ipa/*.ipa` to the entire directory, making the existing cwd assumption destructive. Change to the script directory before cleanup (as `cold-wallet-app/upload_android.sh` already does) or remove an explicitly resolved script-relative path.

Non-blocking metadata note: the PR title says `1.2.0+5`, while `cold-wallet-app/pubspec.yaml`, the PR body, and the described artifacts remain `1.2.0+4`; align whichever version is intended.

Validation at base `f8937d35` / head `68f51b66`:

- Inspected the dependency-fix commit and refreshed complete diff; the dependency manifests and lockfile are restored to their base/canonical graph.
- `dart run melos bootstrap` passed.
- Workspace formatting and `flutter analyze . --fatal-infos` passed; GitHub `Analyze` is green.
- Existing cold-wallet tests passed, plus a temporary focused widget regression covering override -> confirmation -> banner -> re-lock.
- Both upload scripts passed `sh -n`; OpenSSL parsed the new certificate.

`git diff --check` still flags CRLF line endings in the new PEM despite the PR body's clean claim. Android bundle compilation remains unavailable because this reviewer host has no Android SDK. No other blocking findings were found.

n13 (commented):
APPROVE (opinion): both prior blockers are resolved and no blocking findings remain.

The follow-up at `b444d1f5` anchors the broad iOS cleanup to the script's own `mobile-app` directory before `rm -rf build`, and the package/title/body now consistently target `1.2.0+5`. The earlier dependency fix also keeps the canonical dependency graph and removes the Dart 3.12 regression.

Validation at base `f8937d35` / head `b444d1f5`:

- Inspected every follow-up commit and the refreshed complete diff.
- `dart run melos bootstrap` passed.
- Workspace formatting and `flutter analyze . --fatal-infos` passed.
- Cold-wallet `flutter test`: 7/7 passed. The focused override -> confirmation -> banner -> re-lock widget regression passed on the immediately preceding head; the final commit changes only version metadata and the shell-script cwd.
- Both upload scripts passed `sh -n`; OpenSSL parsed the new certificate.
- GitHub `Analyze` passed on this exact head.

Non-blocking caveats: `git diff --check` still reports CRLF endings in the PEM, despite the PR body's clean claim. Android bundle compilation could not be repeated because this reviewer host has no Android SDK; the PR already notes that the final `versionCode 5` artifacts still need rebuilding before distribution.

dewabisma (approved):
LGTM

PR #601, opened 2026-08-11
Cold wallet refinements

n13:
## Summary

- **Re-lock is a normal home-screen button — no overlay.** The connectivity guard now renders nothing while the Wi-Fi lock override is armed; the home screen shows a danger-variant "Online — tap to re-lock" button in its own `bottomContent`, exactly like every other screen button. This replaces two overlay iterations that collided with the app's real UI (a top strip covering the app bar, then a floating bottom button that overlapped the lock screen's Unlock button and needed its own Material ancestor).
- **Change password from Settings.** New Security row opens an action pane: current password (leave empty if none was ever set — the vault happily encrypts under the empty string), new password, confirm. `WalletController.changePassword` verifies the current password by actually decrypting the vault, re-encrypts the mnemonic under the new password with a fresh salt, and re-stores the biometric unlock key when biometrics are enabled (vault re-creation invalidates it by design). Wrong current password, mismatched new passwords, and storage failures each surface distinct errors.
- **Success feedback lives inside the sheet** (checkmark + Done) rather than a SnackBar — `ScaffoldMessenger.showSnackBar` crashed walking a deactivated Scaffold because the lock and guard screens live outside the navigator tree, making root-messenger bookkeeping fragile.

## Validation

- `flutter analyze .` — clean; repo-pinned `dart format --line-length=120` — no changes.
- `flutter test` — cold wallet suite passing.
- Manually exercised on device: override confirm → home shows re-lock button → tap re-arms the guard; password change with empty current password.

### DIFF
--- a/cold-wallet-app/lib/providers/connectivity_provider.dart
+++ b/cold-wallet-app/lib/providers/connectivity_provider.dart
@@ -15,3 +15,9 @@ final networkStatusProvider = StreamProvider<NetworkStatus>((ref) async* {
   yield service.currentStatus;
   yield* service.statusStream;
 });
+
+/// True while any network is reachable. Defaults to true (fail closed) until
+/// the first status arrives.
+final isOnlineProvider = Provider<bool>((ref) {
+  return ref.watch(networkStatusProvider).maybeWhen(data: (s) => s == NetworkStatus.online, orElse: () => true);
+});
--- a/cold-wallet-app/lib/providers/settings_providers.dart
+++ b/cold-wallet-app/lib/providers/settings_providers.dart
@@ -53,7 +53,6 @@ class ColdSettingsController extends Notifier<ColdSettings> {
 
   Future<void> setWifiOverrideEnabled(bool enabled) async {
     state = state.copyWith(wifiOverrideEnabled: enabled);
-    if (!enabled) ref.read(wifiLockOverriddenProvider.notifier).set(false);
     final prefs = await SharedPreferences.getInstance();
     await prefs.setBool(_wifiOverrideKey, enabled);
   }
@@ -72,14 +71,3 @@ class ColdSettingsController extends Notifier<ColdSettings> {
 }
 
 final coldSettingsProvider = NotifierProvider<ColdSettingsController, ColdSettings>(ColdSettingsController.new);
-
-/// Session-only: armed from the connectivity guard after the user confirms the
-/// override warning; disarmed when the persistent toggle is switched off.
-class WifiLockOverridden extends Notifier<bool> {
-  @override
-  bool build() => false;
-
-  void set(bool value) => state = value;
-}
-
-final wifiLockOverriddenProvider = NotifierProvider<WifiLockOverridden, bool>(WifiLockOverridden.new);
--- a/cold-wallet-app/lib/providers/wallet_providers.dart
+++ b/cold-wallet-app/lib/providers/wallet_providers.dart
@@ -1,3 +1,4 @@
+import 'package:cryptography/cryptography.dart' show SecretBoxAuthenticationError;
 import 'package:flutter/foundation.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:quantus_sdk/quantus_sdk.dart';
@@ -5,6 +6,12 @@ import 'package:quantus_cold_wallet/services/cold_auth_service.dart';
 import 'package:quantus_cold_wallet/services/vault_service.dart';
 
 enum WalletStatus { initializing, needsSetup, locked, unlocked }
+
+/// Outcome of a password rotation. The vault write is the commit point:
+/// [changed] and [changedBiometricDisabled] both mean the new password is in
+/// effect; the latter means the biometric unlock key could not be re-stored
+/// and biometric unlock was turned off.
+enum PasswordChangeResult { wrongPassword, changed, changedBiometricDisabled }
 
 @immutable
 class WalletState {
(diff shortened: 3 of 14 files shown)


n13 (commented):
REQUEST_CHANGES (opinion): password rotation can be committed without reliable feedback or a consistent biometric state, and its error boundary currently misreports storage failures.

Findings:

1. **[P1] Do not report a committed vault rotation as failed** — `cold-wallet-app/lib/providers/wallet_providers.dart:102-107` / `cold-wallet-app/lib/services/vault_service.dart:52-54`

   `createVault` writes the newly encrypted vault before deleting the old biometric key. `changePassword` then performs two more fallible storage operations. If the biometric delete, vault re-read, or biometric-key write fails after the vault write succeeds, this method throws and the sheet says "Could not change password" even though the old password has already stopped working. After the normal delete succeeds but re-storing fails, `WalletState.biometricEnabled` also remains `true`, so the next lock screen offers a biometric unlock backed by no usable key. Treat the vault write as an explicit commit point: either make rotation transactional/recoverable, or return/update the committed password and biometric outcome separately so the UI tells the user that the password changed and accurately disables/repairs biometric unlock. Please cover failures at each storage step.

2. **[P1] Keep the password mutation single-flight and visible until it settles** — `cold-wallet-app/lib/widgets/change_password_sheet.dart:12-17,44-56,127`

   `showModalBottomSheet` is dismissible and draggable by default, while the confirm field's `onSubmitted` calls `_submit()` even when `_busy` is already true. The Argon2/storage operation continues after dismissal, but the `!mounted` return discards its result; a user can therefore dismiss what looks like a pending/cancelled action and later discover that the password did change. Repeated keyboard submissions can also start concurrent rotations with separate salts and biometric rewrites. Prevent dismissal/back navigation while the operation is running and enforce `if (_busy) return` (with inputs/submission disabled) so exactly one rotation owns the result.

3. **[P2] Only classify authentication failures as an incorrect current password** — `cold-wallet-app/lib/providers/wallet_providers.dart:96-100`

   `unlockWithPassword` can fail while reading secure storage or decoding a missing/corrupt vault, not only with `SecretBoxAuthenticationError`. Catching every exception here contradicts this method's stated "rethrows storage failures" contract and shows "Current password is incorrect" for device/storage faults. Catch only the authentication-error type; log and rethrow other failures so the sheet can surface the distinct storage error.

Validation:

- Exact base/head: `0f177455fb48265e803f89ddaa056491a376b535...9acab8807b96e1766a269037ad69e2b7fba48e81`
- `dart run melos exec --concurrency=1 -- "flutter analyze . --fatal-infos"` — passed in all four packages
- `dart run melos exec -- dart format lib test --line-length=120 --set-exit-if-changed` — 0 files changed
- `flutter test` in `cold-wallet-app` — 7/7 passed
- `git diff --check` — passed

The existing cold-wallet tests cover transaction signing only; none exercises the new password, biometric-failure, concurrency/dismissal, connectivity-override, or import-error paths.

dewabisma (approved):
LGTM! Nothing blocking just consideration.

n13 (commented):
REQUEST_CHANGES (opinion): the prior auth-classification and concurrency blockers are substantially fixed, but the new pre-commit ordering can silently remove biometric unlock while reporting only a password-change failure.

Findings:

1. **[P1] Surface biometric loss when the vault write fails** — `cold-wallet-app/lib/services/vault_service.dart:52-55`, `cold-wallet-app/lib/providers/wallet_providers.dart:111-117`, `cold-wallet-app/lib/widgets/change_password_sheet.dart:73-78`

   The fix deletes `cold_unlock_key` before writing the replacement vault. If that delete succeeds and the following vault write fails, the old vault/password correctly remain valid, but biometric unlock has been permanently removed. The controller notices this and sets `WalletState.biometricEnabled` to false, then rethrows; the sheet reduces the outcome to `Could not change password: ...` and never tells the user that biometric unlock was also disabled. The next time they lock, the biometric option has disappeared despite the UI only reporting a password-change failure. Return a distinct pre-commit outcome such as `unchangedBiometricDisabled` (or restructure the writes so the old biometric credential survives a failed vault commit), show that outcome explicitly, and add a failure-injection test for delete-success/write-failure.

2. **[P2] Freeze the submitted password fields while rotation is in flight** — `cold-wallet-app/lib/widgets/change_password_sheet.dart:52-59,141-151`

   `_busy` now prevents a second mutation, but all three `PasswordField`s remain editable during the Argon2/storage work. `changePassword` receives the strings captured at the initial call, so edits made while the loader is visible are ignored; on success the form disappears and the user can reasonably believe the final visible new password is the one that was committed. For a credential-changing flow this can make the next unlock unexpectedly fail. Disable or absorb the fields while `_busy` is true so the displayed inputs cannot diverge from the submitted values.

3. **[P2] Log storage failures and keep raw platform errors out of the UI** — `cold-wallet-app/lib/widgets/change_password_sheet.dart:73-78`

   This catch is the terminal handler for pre-commit secure-storage/read/decode failures, but it does not log them and renders `$e` directly. That loses diagnostic evidence and reintroduces the raw-exception UI behavior this PR removed from the import flow. Log the exception and stack trace, then show a stable user-facing storage error.

Resolved from the prior review:

- Only `SecretBoxAuthenticationError` is classified as a wrong password.
- A committed password change with biometric re-store failure now has a typed, visible outcome and updates `WalletState`.
- `_busy`, `PopScope`, and disabled drag make the mutation single-flight and prevent dismissal while it runs.

Validation:

- Exact base/head: `0f177455fb48265e803f89ddaa056491a376b535...14c108106428ecb7075fc735e7f8ca9e3b8af405`
- Intervening commit inspected: `14c108106428ecb7075fc735e7f8ca9e3b8af405`
- `flutter test` in `cold-wallet-app` — 11/11 passed, including 4 new password tests
- `dart run melos exec --concurrency=1 -- "flutter analyze . --fatal-infos"` — passed in all four packages
- `dart run melos exec -- dart format lib test --line-length=120 --set-exit-if-changed` — 0 files changed
- `git diff --check` — passed
- GitHub `Analyze` — passed on the current head

The added tests cover successful rotation, empty/current wrong-password behavior, and successful biometric re-keying, but not the storage-failure outcomes or busy-sheet interaction above.

n13 (commented):
REQUEST_CHANGES (opinion): the explicit findings from the last review are fixed, but password rotation still leaves the vault and biometric credential crash-inconsistent.

Finding:

1. **[P1] Detect and clear a biometric key that belongs to the previous vault** — `cold-wallet-app/lib/services/vault_service.dart:34-36,44-55`, `cold-wallet-app/lib/providers/wallet_providers.dart:45-52,113-131`

   `createVault` now commits the replacement vault while the old `cold_unlock_key` remains in secure storage; the controller only overwrites or deletes that key afterward. If the process is terminated after the vault write but before either operation completes, no catch/cleanup runs. The same persistent state occurs when biometric re-store and the attempted cleanup delete both fail. On the next launch, `isBiometricEnabled()` checks only that a key exists, so `_init` marks biometrics enabled and the lock screen keeps offering biometric unlock even though that stale key can never decrypt the new vault.

   I reproduced the persisted state with the storage mock: create an `alpha` vault and biometric key, write a `beta` vault without touching the biometric entry (the exact interrupted interval), then `isBiometricEnabled()` returns `true` while `unlockWithBiometricKey()` throws `SecretBoxAuthenticationError`.

   Make the vault/key relationship self-validating across process death—for example, version both entries and require matching versions, or validate the stored biometric key against the current vault during initialization and remove/ignore it on authentication mismatch. Add an interrupted-rotation regression test using a fresh controller/container so startup reports biometrics disabled instead of persisting a broken unlock option.

Resolved from the prior review:

- A failed vault write preserves both the old password and old biometric key, with injected-failure coverage.
- A failed biometric re-store produces the typed disabled outcome and attempts stale-key cleanup.
- Password fields are frozen while rotation is in flight, with widget coverage.
- Terminal storage errors are logged with stack traces and shown through stable user-facing text.

Validation:

- Exact base/head: `0f177455fb48265e803f89ddaa056491a376b535...f8ce9ef93fe99f3cc39acc82f373d95ccdd37cd7`
- Intervening commit inspected: `f8ce9ef93fe99f3cc39acc82f373d95ccdd37cd7`
- `flutter test` in `cold-wallet-app` — 14/14 passed
- Focused interrupted-rotation probe — reproduced stale `isBiometricEnabled() == true` plus biometric decryption failure
- `dart run melos exec --concurrency=1 -- "flutter analyze . --fatal-infos"` — passed in all four packages
- `dart run melos exec -- dart format lib test --line-length=120 --set-exit-if-changed` — 0 files changed
- `git diff --check` — passed
- GitHub `Analyze` — passed on the current head

n13 (commented):
APPROVE (opinion): the crash-consistency blocker is fixed and I found no remaining blocking issues.

The new biometric entry pairs the derived key with the current vault salt. After an interrupted password rotation, a fresh controller now detects the generation mismatch, removes the orphaned key, and initializes with biometric unlock disabled while preserving password access to the committed vault. `unlockWithBiometricKey` also removes a pre-pairing legacy key if authentication proves it stale, and the controller immediately stops offering that path. This closes the prior persisted broken-unlock state without decrypting a valid legacy vault during startup.

The earlier findings remain resolved: password rotation has explicit pre/post-commit outcomes, only authentication failures are classified as a wrong password, failed writes preserve the old credentials, biometric re-store failures are reported honestly, the sheet is single-flight/non-dismissible while busy, fields freeze after submission, and storage failures are logged behind stable UI text.

Validation:

- Exact base/head: `0f177455fb48265e803f89ddaa056491a376b535...92d3f9c213fe00857929b4a33b37e6f054334a02`
- Intervening commit inspected: `92d3f9c213fe00857929b4a33b37e6f054334a02`
- `flutter test` in `cold-wallet-app` — 16/16 passed, including fresh-controller interruption recovery, legacy-key self-healing, injected storage failures, and busy-sheet coverage
- `dart run melos exec --concurrency=1 -- "flutter analyze . --fatal-infos"` — passed in all four packages
- `dart run melos exec -- dart format lib test --line-length=120 --set-exit-if-changed` — 0 files changed
- `git diff --check` — passed
- GitHub `Analyze` — passed on the current head

No blocking findings remain.

PR #602, opened 2026-08-12
Cold wallet: fix Android camera, rebuild the signing screen (1.2.0+7)

n13:
# Human Overview

- Fixes obscure Android camera bug - camera would just not work, even though it worked on iOS. Fixed now, see below.
- Vastly improves the multisig and send and vote displays on signing, removing tons of unneeded fields and improving legibility by a lot, screens below.

## Android camera — release builds only

`Camera unavailable … (genericError)` on Android device installs, while iOS, debug-over-cable and the hot wallet all worked.

Root cause is R8, not the scanner code. AGP shrinks release builds by default, and the cold wallet had no `minifyEnabled` setting of its own. ML Kit's consumer rule is:

```
-keep class * implements com.google.firebase.components.ComponentRegistrar
```

R8 full mode honours that by keeping the class but **dropping the no-arg constructor** — which is exactly how ML Kit instantiates it, reflectively, from the `ComponentDiscoveryService` manifest entries. `BarcodeScanning.getClient()` then throws, the exception escapes `MobileScannerHandler.start`, and mobile_scanner surfaces it as `genericError`.

Proof from the previous release build's own R8 output:

```
# usage.txt (removed by R8)
com.google.mlkit.vision.barcode.internal.BarcodeRegistrar:
    public void <init>()

# mapping.txt (renamed)
BarcodeRegistrar -> BarcodeRegistrar:
    java.util.List getComponents() -> a
```

The hot wallet escapes this only because its `gradle.properties` carries `android.r8.strictFullModeForKeepRules=false`.

**Fix:** release builds skip shrinking. The Dart code is AOT-compiled and only the thin plugin layer is Java, so shrinking buys this app nothing and costs a whole class of release-only reflection breakage across the crypto, secure-storage and biometric plugins too.

Verified on the 1.2.0+7 APK: `BarcodeRegistrar` and `getComponents` are present under their real names, and no `mapping.txt` is produced. The shrink settings are unchanged since.

The error panel was also a dead end — it blamed permissions for every failure code, including this one. It now reports the actual error and offers a retry; a retry that throws instead of re-rendering the panel reports in the panel too.

## Review & Sign

The screen led with signed extensions (genesis hash, era, tip, metadata hash) and pushed the amount and recipient off the top of the screen.

**Send** now reads: `SEND` → the amount, largest thing on screen → `TO` address + checkphrase → `FROM` address + checkphrase. Fits one screen; Sign is reachable without scrolling.

The headline is deliberately opinionated: it names the one distinction a signer actually weighs — `SEND`, `REVERSIBLE SEND`, `ASSET SEND`, `REVERSIBLE ASSET SEND` — because `transfer_allow_death` vs `transfer_keep_alive` is not a user-relevant difference. A reversible send keeps its reversibility window on screen as the Delay row.

**Multisig approve** reads: `MULTISIG APPROVE` → the boxed call being approved (`SEND`, amount, `TO`, address, checkphrase) → `SIGNED BY` → multisig account and proposal id.

Signed extensions and the raw payload move behind an **Advanced** disclosure, listed plainly.

Nothing is hidden that a signer needs. The summary is a lead, not a replacement: any call parameter the amount/recipient pair does not already show is still listed underneath, matched **by identity rather than by label or value** (`TransferSummary` names the exact fields it restates), so neither a renamed parameter nor a second parameter that happens to carry the same value can be silently dropped — a `force_transfer` whose source equals its destination still shows its `Source` row. A `schedule_transfer_with_delay` still shows its delay; a wrapper still shows its own parameters.

The signer's own row claims `From` only when the summary shows a plain send and the call names no other account the funds could leave instead; a `force_transfer` or a no-value call reads `Signed by`.

### Worth a look
- **The scroll-to-enable gate on Sign is removed.** It forced a read of a parameter list that no longer exists, and it was implemented with a post-frame callback. Say the word if you want it back for the multisig case, which still scrolls.
- `Multisig · approve`'s nested call is labelled `You are approving` (proposals: `You are proposing`).
- Headlines never show runtime naming, at any depth. The exact `pallet · call` identity of the whole payload — nested calls included — is one line in the Advanced sheet: `Call: Multisig · approve → Balances · transfer_allow_death` (`DecodedCall.displayTitleChain`). Comments on `actionTitle`, `CallDetailView` and the headline state that this collapse is intentional, so it stops getting flagged as information loss.

### DRY
- `_AddressField` moves out of `call_detail_view.dart` into a shared `AddressWithCheckphrase`, now used for the signer too — one code path for every address on screen.
- `DecodedCall.actionTitle` lives in the SDK next to the model. Only a call that moves value *itself* reads as `SEND`; a wrapper carries the inner summary but names itself, so an approval can never masquerade as a plain send. The "is this a wrapper" check is a single `DecodedCall.isWrapper`, shared by the title, the hero summary and the screen, and it sees a nested call even inside a parameter group.
- `checkphraseProvider` is the shared `addressCheckphraseProvider` family applied to the wallet's own address.

## Tests
`melos run format` / `melos run analyze` clean. Cold wallet 26/26, SDK decoder + parser suites 46/46. New `call_display_test.dart` pins the screen's contract: every `CallField` reaches the screen exactly once (own row or summary hero), headlines summarise while the exact call chain is reachable under Advanced, a reversible send names itself and shows its window, and the signer row's `From`/`Signed by` labels — covering plain sends, `force_transfer` with `source == dest`, multisig approve and a no-value vote. SDK `actionTitle` cases cover send, reversible/asset kinds, wrapper-never-SEND, the no-value fallback, and the display-title chain.

## Build
Bumped to **1.2.0+8**. Release AAB and APK built locally from the 1.2.0+7 config and verified unminified; shrink settings unchanged since. No Play service-account key on this machine, so the internal-track upload (`cold-wallet-app/upload_android.sh --upload`) still needs to be run by someone who has it.


n13 (commented):
The Android/R8 diagnosis is excellent and the new screen reads far better. Two things in the rebuild remove information a signer needs, though, and I don't think they were intended — the PR description says one of them is still there.

## Blocking

**1. The exact `pallet · call` is no longer on screen — `sign_transaction_screen.dart:138-141`, `call_detail_view.dart:118`**

`actionTitle` collapses every value-moving call to the single word `SEND`: `Balances.transfer_allow_death`, `Balances.transfer_keep_alive`, `Balances.force_transfer`, `ReversibleTransfers.schedule_transfer`, `schedule_transfer_with_delay` and `Assets.transfer` all render identically in the headline. The only place `displayTitle` survives is inside the Advanced sheet (`sign_transaction_screen.dart:229`).

The description says *"The precise `pallet · call` stays as a small muted line under the headline, so `transfer_allow_death` vs `transfer_keep_alive` is still verifiable."* There is no such line in the code — the old `_hero` rendered it, and it went away with `_hero`. A signer looking at `SEND / 10 QUAN / To … / From …` cannot tell whether the call may reap their account, or whether it is an immediate transfer at all, without tapping Advanced.

**2. A nested call's identity is nowhere at all — `call_detail_view.dart:118` + `sign_transaction_screen.dart:229`**

For `Multisig · approve` wrapping `Balances · transfer_allow_death`, the inner box is headed by `inner.actionTitle` → `SEND`, and `_advancedLines` lists only the **outer** `parsed.call.displayTitle`. So the pallet and call actually being authorised appear on no screen and in no sheet. Before this PR, `CallDetailView` printed `displayTitle` at every depth.

That contradicts the contract this file is built on — `decoded_call.dart:8-11` ("nothing is summarised away") and the class doc you kept at `call_detail_view.dart:99-100` ("if a byte is being signed, it is on this screen"). The call index bytes are being signed.

Cheapest fix for both: keep `displayTitle` as the muted sub-line under `actionTitle` in `CallDetailView`, and render the top-level one the same way — that also makes the description accurate again.

## Worth fixing

**3. `coveredBySummary` suppresses by value, so a duplicate-valued parameter is silently dropped — `call_detail_view.dart:12-18`**

The doc argues value-matching means a *renamed* parameter can never be dropped, which is true — but the converse isn't handled: any field that merely *equals* the summary is dropped, regardless of its own label. `Balances.force_transfer(source: X, dest: X, value: V)` (`call_decoder.dart:98-105`) → `summary.recipient == X`, so **both** the `Source` and `Destination` rows disappear and the screen shows only `To X`; the signer never learns a `Source` parameter exists. Suppressing at most one field per summary component (or matching by field identity rather than value) closes it.

**4. `From: <signer>` is asserted whenever there is no nested call — `sign_transaction_screen.dart:186-189`**

The label is derived from "has no nested call", not from whether the signer is actually the source. For `force_transfer` the screen reads `SEND / V / To: dest / Source: X / From: <signer>` — funds leave `X`, not the signer. For a call that moves nothing (`Timestamp · set`) a bare `From` row implies a transfer that isn't happening. Consider deriving the label from `summary != null && recipient != null` instead.

**5. DRY — "is this a wrapper call" now exists in three copies**

`decoded_call.dart:125`, `call_detail_view.dart:26`, `sign_transaction_screen.dart:186`. Per the DRY rule in CLAUDE.md ("If existing code can't be used directly, abstract out the shared parts and use them in both places"), this wants `bool get isWrapper => fields.any((f) => f is NestedCallField)` on `DecodedCall`, used by all three. Not cosmetic: `CallDecoder` can already emit a `NestedCallField` inside a `FieldGroup` (`call_decoder.dart:887`, reachable from the generic path at `:801`/`:808`), and all three copies miss that shape — the day such a call carries a summary, the headline reads `SEND` for a wrapper unless every copy is updated in lockstep.

**6. DRY — `addressCheckphraseProvider` duplicates `checkphraseProvider`**

`address_with_checkphrase.dart:9-11` vs `wallet_providers.dart:114-119` — same body; the latter is just the former applied to `addressProvider`. Now that the family version is public, rewrite `checkphraseProvider` in terms of it.

**7. No test covers the logic that decides what reaches the screen**

The premise of the rebuild is "nothing a signer needs is hidden", and `coveredBySummary` / `callSummaryBody` / `_callBody` have no tests — `cold-wallet-app/test/` still holds only `transaction_signing_test.dart`, touched here for a label rename. A test asserting that for each decoder fixture every `CallField` appears exactly once (via the summary or as its own row) would have caught 2, 3 and 4.

## Minor

**8. `_restartCamera` silently does nothing on the throwing paths — `scan_transaction_screen.dart:119-125`**

The comment is right that `start()` folds `MobileScannerException` into `value.error` (and `MobileScannerState.copyWith` passes `error: error`, not `error ?? this.error`, so a successful retry does clear the panel — good). But `start()` *throws* for `controllerDisposed`, `controllerNotAttached` and `controllerInitializing`. In those cases the tap produces no visible change whatsoever, only a `debugPrint`. CLAUDE.md: "Fail early — never allow a silent failure." Surface it in the panel.

**9. Disabling R8 wholesale is wider than the diagnosed cause — `build.gradle.kts:60-61`**

`mobile-app/android/gradle.properties` already carries `android.r8.strictFullModeForKeepRules=false` — the one-line escape your own writeup identifies as why the hot wallet is unaffected. Setting it here keeps shrinking and obfuscation on for a wallet binary and leaves the two apps configured the same way. The "Dart is AOT, shrinking buys nothing" argument is defensible, so this is a judgement call, not a bug — just worth making deliberately rather than by omission.

**10. The description no longer matches the head commit**

Version is `1.2.0+8`, not `+7`; `CALL BEING APPROVED` was renamed to `You are approving` / `You are proposing` (`call_decoder.dart:335,350`) despite the note saying it stays; and the `pallet · call` line is gone (item 1). The "verified on the 1.2.0+7 APK" evidence therefore describes an earlier build than this diff.

---

**Verdict: request changes** — items 1 and 2 remove the exact call identity from a signing screen whose contract is that it hides nothing; the rest are non-blocking.

n13 (commented):
Verdict: **REQUEST CHANGES** (posted as a comment because this is the authenticated account's own PR). The original nested-call identity, duplicate-field suppression, and signer-label findings are addressed, but two new signing-display regressions remain at `9102e20a`.

## Blocking

1. **Explicit timestamp windows are now displayed lossily — `quantus_sdk/lib/src/chain/call_decoder.dart:892-900`**

   `_delayField` used to keep the exact millisecond value next to the friendly duration; the new code keeps only `DatetimeFormattingService.formatDuration(...).formatted`. That formatter always drops milliseconds and, once the duration reaches a day, drops seconds too. Consequently two distinct signed calls such as `Timestamp(86_400_000)` and `Timestamp(86_424_000)` both render as `01d:00h:00m`, even though their encoded reversal windows differ. An air-gapped signer cannot distinguish those payloads from the review screen. Keep an exact value (the previous `($millis ms)` is sufficient), or make the friendly formatter lossless, and add a regression test using two colliding durations instead of asserting that `ms` is absent.

2. **The asset hero suppresses the only raw-unit/decimals caveat — `cold-wallet-app/lib/components/call_detail_view.dart:36-40,58-65,156-168`**

   `callSummaryBody` renders `TransferAmount` and suppresses the matching `AmountField`. For assets, `TransferAmount` now says only `4200 units of asset 1`; the explanation that asset decimals are not in the payload and the amount is raw exists only in `CallFieldView`, which is precisely the widget suppressed on this path. This makes a smallest-unit integer look like a human asset quantity and contradicts the stated intent that the decimals caveat remains visible. Put `raw` back in the hero or render the caveat beneath it, with a widget test that asserts the caveat is actually on screen.

## Non-blocking

- **A recovered camera can retain a stale failure banner — `cold-wallet-app/lib/screens/scan_transaction_screen.dart:116-122`**. After one throwing retry sets `_error`, a later successful `_controller.start()` never clears it, so scanning can resume while `Camera restart failed` remains in red. Clear `_error` when beginning or completing a successful retry.

## Validation

- `git diff --check 1072bf50...9102e20a` — passed.
- `dart run melos exec -- dart format lib test --line-length=120` — passed, 0 files changed.
- `dart run melos exec --concurrency=1 -- "flutter analyze . --fatal-infos"` — all 4 packages passed.
- Cold-wallet changed suites — 17/17 passed.
- SDK decoder/parser suites — 46/46 passed.
- GitHub `Analyze` — passed.

The direct pinned Melos commands were used because `dart run melos run format` shells out to an unavailable global `melos` executable in this checkout.

PR #606, opened 2026-08-14
fix ZK proofs

n13:
## Summary
- **align Wormhole claim proofs with the live 4 bps runtime fee and derive miner/mobile fee labels from the proof constant**
- isolate generated v4.2.0 circuit artifacts in app-support storage and validate the versioned manifest before proving
- route both claim consumers through the shared SDK circuit manager
- remove obsolete circuit-manager APIs and unused Rust fee/asset constants
- add regression coverage for circuit versioning, config validation, and the live fee calculation

## Diagnosis
The screenshot failure is not a Wormhole call-shape or prover-version mismatch. The live runtime is spec 144 / transaction version 3, and both the app and chain use Wormhole circuits 4.2.0. The encoded call remains Wormhole pallet index 20 and verify_private_batch call index 2.

The app was still committing volume_fee_bps = 10 into every proof while the runtime requires 4. Runtime pre-validation rejects that public input before proof verification and maps the rejection to InvalidTransaction::Call, which surfaces through author_submitExtrinsic as Transaction call is not expected.

## Verification
- melos run format
- melos run analyze
- cargo fmt --check in quantus_sdk/rust
- flutter test test/services/wormhole_coin_selection_test.dart test/services/wormhole_utxo_service_test.dart test/services/circuit_manager_test.dart test/services/wormhole_send_service_test.dart (18 passed)

### DIFF
--- a/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart
+++ b/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart
@@ -307,7 +307,7 @@ class _ClaimRewardsDialogState extends State<_ClaimRewardsDialog> {
           const SizedBox(height: 12),
           _confirmRow('Destination', _addressController.text.trim(), mono: true),
           const SizedBox(height: 12),
-          _confirmRow('Fee', '0.1% volume fee'),
+          _confirmRow('Fee', '${wormholeVolumeFeePercentText()}% volume fee'),
           const SizedBox(height: 24),
           Container(
             padding: const EdgeInsets.all(12),
--- a/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart
+++ b/miner-app/lib/features/withdrawal/claim_rewards_dialog.dart
@@ -1,7 +1,4 @@
-import 'dart:io';
-
 import 'package:flutter/material.dart';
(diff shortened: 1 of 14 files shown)


n13 (commented):
APPROVE — no blocking findings at `53ead08d`.

The Wormhole proof input and per-leaf net calculation now use the Chain runtime's 4 bps fee, and both miner/mobile fee labels derive from that same value. The shared circuit manager preserves the Rust API's base-directory contract, resolves the `v4.2.0` artifact directory consistently, and fails on missing or invalid manifest data. All claim/send consumers reach that shared path.

Validation:
- `git diff --check 3675320e...53ead08d`
- `dart run melos exec -- dart format lib test --line-length=120 --set-exit-if-changed` — 0 changes
- repository-wide `flutter analyze . --fatal-infos` — clean in all 4 packages
- `cargo fmt --check` — passed
- `cargo test --locked` — 40 passed
- `cargo build --release --locked` — passed
- focused Flutter suite for coin selection, UTXO, circuit manager, and send service — 18 passed with the rebuilt native library
- GitHub `Analyze` check at `53ead08d` — passed

No blocking findings.

dewabisma (approved):
LGTM!

Just one suggestion if possible.

PR #613, opened 2026-08-19
Update design system

dewabisma:
## Summary
- Extract repeating widget as reusable components in SDK
- Add v3 design tokens
- Add new components
- Update some existing component design tokens
- Update cold wallet theme
- Partially update mobile app theme


n13 (changes requested):
**Reviewer model:** GPT Sol

REQUEST_CHANGES — the shared design-system extraction introduces two user-visible regressions, and the new cold-wallet suite does not pass.

Blocking findings:

1. `quantus_sdk/lib/src/ui/themes/app_theme.dart:41` no longer installs an `AppSizeTheme` extension. `mobile-app/lib/v2/screens/auth/auth_wrapper.dart:54` still evaluates `context.themeSize.screenPadding` on the unauthenticated, non-authenticating path, while the getter null-asserts `Theme.of(context).extension<AppSizeTheme>()`. The wallet therefore throws while building its unlock screen. Add the appropriate default/iPad `AppSizeTheme` to the shared theme extensions and cover the unauthenticated AuthWrapper state with a widget test.

2. `quantus_sdk/lib/src/ui/components/scaffold_base.dart:52` changed the previously mobile-specific scaffold from always rendering `NetworkStatusBanner` to rendering only an optional `networkBanner`. None of the mobile app's ScaffoldBase call sites supplies that argument, and the moved `mobile-app/lib/v2/components/network_status_banner.dart` has no consumer. As a result, the offline "Not Connected" warning disappears from every mobile screen. Preserve the mobile wrapper that injects the banner, or mount the banner once at the mobile app level.

3. `cold-wallet-app/test/qr_tuning_controls_test.dart:41` deterministically expects `15 FPS` and `1100 bytes` after seeding empty preferences, but `ColdSettings` defaults are `10` and `300`. The lookup at line 42 throws `Bad state: No element`; the targeted test failed in three consecutive runs and the full cold-wallet suite finished with 39 passes and this failure. Seed the intended values or assert the actual defaults, and add the cold-wallet suite to CI so this cannot remain hidden behind the current green job.

Validation:

- Reviewed exact head `80f8279fb02578fa9d5f3e9b6934970f7f7cd079` against base `0659cf818b1a226a83f5b56217b709a95a1b6646`.
- GitHub CI passed formatting, analysis, 288 mobile tests, and 258 non-native SDK tests, but does not run cold-wallet tests.
- Workspace Dart formatting check passed with 0 changes.
- The other 39 cold-wallet tests passed when the failing QR tuning test was excluded.
- `git diff --check` also reports trailing whitespace at `quantus_sdk/pubspec.yaml:45` and `:55`.

n13 (approved):
**Reviewer model:** GPT Sol

APPROVE — the three prior blockers are resolved at the current head, and I found no remaining correctness or compatibility issue.

The mobile-specific `ScaffoldBase` wrapper now injects `NetworkStatusBanner` for both static and refreshable layouts, and every mobile `ScaffoldBase` consumer routes through that wrapper. The cold-wallet QR-tuning test now asserts the real settings defaults and the full cold-wallet suite is included in CI. `AuthWrapper` no longer dereferences the removed `AppSizeTheme`. I also rechecked the shared decoded-call/signing display and theme/component moves across the complete diff.

Validation:

- Reviewed exact head `fe597d11010005e0ce5846752c2a1e2ec81265cf` against base `0659cf818b1a226a83f5b56217b709a95a1b6646`, including the four commits since the prior review.
- GitHub CI passed on the current head.
- Workspace bootstrap passed.
- Dart formatting check passed across all four packages with 0 changes.
- Flutter analysis reported no issues across cold wallet, miner, SDK, and mobile; analysis processes were stopped or checked at the required 10-second boundary.
- Cold wallet: 40 tests passed.
- Quantus SDK: 258 non-native tests passed.
- Mobile app: 291 tests passed, including the restored offline-banner regression tests.

Non-blocking cleanup: `git diff --check` still reports trailing whitespace in `quantus_sdk/pubspec.yaml:45` and `:55`.

No blocking findings.

PR #615, opened 2026-08-21
fix: scope hide-balances to the home screen

n13:
## Problem

`isBalanceHiddenProvider` was read *inside* the amount formatters — `balanceDisplayProvider` and `txAmountDisplayProvider` in `currency_display_provider.dart`. Every screen that formats an amount goes through that provider, so the eye toggle on home masked amounts that have nothing to do with the account balance:

- multisig open proposals (`ProposalListTile`) and decoded calls in the approve/execute/cancel confirm sheets
- POS charge + QR screens
- send input amount, review send
- transaction detail sheet

## Change

The formatter no longer has any notion of hidden balances. The home screen owns the flag and passes it explicitly to the components it contains:

- `AmountDisplayWithConversion.isHidden` → masks the main balance
- `ActivitySection.isHidden` / `MultisigActivitySection.isHidden` → masks the amounts in the home transaction list

Everything else formats amounts normally, always. Open multisig proposals are **not** masked even on a multisig home — they are pending decisions to review, not balance disclosure.

The toggle itself still persists via `SettingsService`; only the home screen reads it now.

## DRY

- `txItemAmountText(data, format, {isHidden})` in `tx_item.dart` replaces three copies of the `hideAmount ? '—' : format(...).primaryAmount` ternary (home, multisig, activity screen)
- the formatter's inline function type is now a `TxAmountFormatter` typedef
- `customHiddenText` is gone — its only caller passed the same `'-----'` the default already used; the placeholder now lives once, next to the component that renders it
- `_toFiatDisplayState` lost two parameters and its post-hoc mutation, so it just builds and returns the state

## Verification

- `dart analyze lib test` in `mobile-app`: no issues
- `flutter test` in `mobile-app`: 291 tests pass
- `melos run format` clean

(`melos run analyze` can't complete locally — it dies in `flutter pub get` on a firebase_messaging plugin rsync error unrelated to this change.)

### DIFF
--- a/mobile-app/lib/providers/currency_display_provider.dart
+++ b/mobile-app/lib/providers/currency_display_provider.dart
@@ -239,16 +239,17 @@ class CurrencyDisplayState {
 // Balance display provider
 // ---------------------------------------------------------------------------
 
-final _hiddenAmountText = '-----';
-
-/// Combines balance, hidden state, flip state, selected fiat, and exchange
-/// rate into [CurrencyDisplayState] ready for widgets to render.
+/// Combines balance, flip state, selected fiat, and exchange rate into
+/// [CurrencyDisplayState] ready for widgets to render.
+///
+/// Hiding balances is a UX concern owned by the screen that offers the toggle,
+/// which passes it into AmountDisplayWithConversion; it is deliberately absent
+/// here so unrelated screens never inherit it.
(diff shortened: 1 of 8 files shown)


dewabisma (approved):
Make sense to only hide in home. Just wondering if user ever want show activity history screen with balance hidden? Maybe not.

LGTM!

PR #617, opened 2026-08-24
APK signing improvement for cold wallet

n13:
## Changes

- `build_wallet_apk.yml` declares its signing inputs as `workflow_call` secrets (`KEYSTORE_BASE64`, `STORE_PASSWORD`, `KEY_ALIAS`, `KEY_PASSWORD`, plus the mobile-only `MOBILE_APP_ENV` / `GOOGLE_SERVICES_JSON`).
- `create_wallet_android_build.yml` passes `ANDROID_*` to the mobile job and `COLD_ANDROID_*` to the cold job instead of `secrets: inherit`.
- A `Check required secrets` step fails the build, naming the secret, if any required one is empty.

### DIFF
--- a/.github/workflows/build_wallet_apk.yml
+++ b/.github/workflows/build_wallet_apk.yml
@@ -13,6 +13,19 @@ on:
         description: "Build mode"
         required: true
         type: string
+    secrets:
+      KEYSTORE_BASE64:
+        required: true
+      STORE_PASSWORD:
+        required: true
+      KEY_ALIAS:
+        required: true
+      KEY_PASSWORD:
+        required: true
+      MOBILE_APP_ENV:
+        required: false
+      GOOGLE_SERVICES_JSON:
+        required: false
     outputs:
       version:
         description: "pubspec version (e.g. 1.5.10+127)"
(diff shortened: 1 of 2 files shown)


dewabisma (commented):
Why we make it different between two workflow? do we have like 3 set of signing secrets????

n13 (commented):
Verdict: LGTM — no blocking findings.

There are two signing-secret sets, not three: ANDROID_* for the mobile wallet and COLD_ANDROID_* for the cold wallet. create_wallet_android_build.yml maps the selected set into the generic KEYSTORE_BASE64 / STORE_PASSWORD / KEY_ALIAS / KEY_PASSWORD contract of build_wallet_apk.yml. That keeps each reusable-workflow job scoped to the one keyset it needs and removes the internal app-specific secret selection that caused the original mix-up.

I also checked the full diff, linted both changed workflows, and confirmed the Analyze check is green. The required-secret check fails before the expensive build steps and names the missing mapped secret.

dewabisma (approved):
Ah... Okay, after reading the whole workflow I get it. LGTM!

